PR #19, opened 2016-11-18
Changes the response format for api requests

Mironor:
@alangorton AddressResponse.scala is what you want
There are some hard-coded values (like limit or offset), I'd like to leave them as they are before we know what to do with them.


server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,251 @@
...
+      other.welshDependentLocality,
+      other.welshPostTown,
+      other.poBoxNumber,
+      other.startDate,
+      other.endDate
+    )),
+    nag = None,
+    geo = None,
+    underlyingScore = 1,
+    underlyingMaxScore = 1
```
Mironor: Those 2 numbers are just stubs right now, they will be changed in future PRs

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
```diff
@@ -42,54 +43,78 @@ class AddressController @Inject()(esRepo : ElasticsearchRepository)(implicit ec
...
+    Future.fromTry(format.toAddressScheme()).flatMap {
+      case PostcodeAddressFile(str) => esRepo.queryAddress(tokens).map {
+        addresses => addressResponse(tokens, addresses)
+      }
       case _ => asyncWrongFormatResponse
     }
   }
 
+  private def addressResponse(tokens: AddressTokens, addresses: Seq[PostcodeAddressFileAddress]): Result = {
+    val response = AddressBySearchResponseContainer(
```
Mironor: Why?

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
```diff
@@ -42,54 +43,78 @@ class AddressController @Inject()(esRepo : ElasticsearchRepository)(implicit ec
...
-    logger info "#addressQuery got a response from es"
-    Ok(
-      Json toJson(
-        addresses map PostcodeAddressFileReplyUnit.fromPostcodeAddressFileAddress
-      )
-    )
-  }
 
-  private val asyncWrongFormatResponse : Future[Result] = Future successful Ok("Wrong format")
+  private val asyncWrongFormatResponse: Future[Result] = Future.successful(Ok("Wrong format"))
```
Mironor: I think this is even better to be changed to AddressByUprnResponseFormat with Option for the address and an error in the errors array.

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,251 @@
...
+  message: String
+)
+
+/**
+  * Contains one response error
+  *
+  * @param code    error code
+  * @param message error description
+  */
+case class AddressResponseError(
```
Mironor: You mean to replace it with a Status? Aren't these 2 entities have different meaning?

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
```diff
@@ -1,44 +1,118 @@
...
     override def createAll(): Future[Seq[_]] = ???
+
     override def deleteAll(): Future[Seq[_]] = ???
-    override def queryAddress(buildingNumber: Int, postcode: String): Future[Seq[PostcodeAddressFileAddress]] = ???
+
+    override def queryAddress(tokens: AddressTokens): Future[Seq[PostcodeAddressFileAddress]] =
+      Future.successful(Seq(validPafAddress))
+
+
     override def client(): ElasticClient = ???
```
analytically: Null??

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
Mironor: And a different companion object with different predefs

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
Mironor: `ElasticClient.local(Settings.builder().build())` instead of null

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,251 @@
...
+
+import play.api.libs.json.Json
+import uk.gov.ons.addressIndex.model.db.index.PostcodeAddressFileAddress
+
+/**
+  * Contains implicit readers/writers to convert response models into json (play standard).
+  * When needed, do following:
+  * `import uk.gov.ons.addressIndex.server.model.response.implicits._`
+  */
+object implicits {
```
alangorton: Use Double instead of Float for latitude / longitude. Same for Elastic scores?

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
iRhysBradbury: Why not
```
case class Status(
  message: String,
  code: 
)

object AddressError extends Status(
 message = "My AddressError Message",
 code = 505
)
```

iRhysBradbury: You can also have the PlayJson formats on the status companion

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
iRhysBradbury: +1 @Mironor

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
iRhysBradbury: `rootMaxScore`
`rootScore`

same meaning, less letters :)

Mironor: @alangorton was the one to choose those names, Alan, why "underlying" ?

alangorton: These will be set to the raw Elastic scores for demonstration purposes. It's extremely likely that we'll replace these two fields with something else - heck, we might even rename them to _underlyingXxx and add two new fields score and maxScore. I wouldn't worry about it. 'Root' makes them sound more important than they actually are.

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
alangorton: Let's avoid inheritance in this case. It's an unnecessary coupling..

iRhysBradbury: The gains would be quite nice actually. You could return that case object as the response if you extended it this way.

model/src/main/scala/uk/gov/ons/addressIndex/model/AddressIndexModel.scala
```diff
@@ -16,11 +15,11 @@ case class AddressIndexSearchRequest(
 )
 
 sealed trait AddressScheme {
-  override def toString() : String
+  override def toString: String
 }
 case class PostcodeAddressFile(override val toString : String) extends AddressScheme
 case class BritishStandard7666(override val toString : String) extends AddressScheme
-case class InvalidAddressSchemeException(str : String) extends IllegalArgumentException(s"Invalid address scheme $str")
+case class UnsupportedScheme(override val toString : String) extends AddressScheme
```
iRhysBradbury: How can an UnsupportedScheme be a type of AddressScheme

iRhysBradbury: why? This doesnt make any sence. How can an `UnsupportedScheme` be a type of `AddressScheme`

model/src/test/scala/uk/gov/ons/addressIndex/model/AddressIndexModelTest.scala
```diff
@@ -7,34 +7,27 @@ class AddressIndexModelTest extends FlatSpec with Matchers {
...
   it should "produce the correct `AddressScheme` (`PostcodeAddressFile`) for the given string `postcodeAddressFile`" in {
     val sample = "postcodeAddressFile"
-    sample.toAddressScheme map (_ shouldBe a [PostcodeAddressFile])
+    sample.stringToScheme shouldBe a [PostcodeAddressFile]
   }
 
   it should "produce the correct `AddressScheme` (`BritishStandard7666`) for the given string `britishstandard7666`" in {
     val sample = "britishstandard7666"
-    sample.toAddressScheme map(_ shouldBe a [BritishStandard7666])
+    sample.stringToScheme shouldBe a [BritishStandard7666]
```
iRhysBradbury: why is this stringToScheme? All implicit classed from are implicit in their usage, and all to's are in the defs.
I dont know why this has been changed, considering it didn't need to AND how it was previously is using the naming convention of the scala language itself.

model/src/test/scala/uk/gov/ons/addressIndex/model/AddressIndexModelTest.scala
```diff
@@ -7,34 +7,27 @@ class AddressIndexModelTest extends FlatSpec with Matchers {
...
     val sample = "bs7666"
-    sample.toAddressScheme map(_ shouldBe a [BritishStandard7666])
+    sample.stringToScheme shouldBe a [BritishStandard7666]
   }
 
   it should "produce the correct `AddressScheme` (`BritishStandard7666`) for the given string `bs`" in {
     val sample = "bs"
-    sample.toAddressScheme map(_ shouldBe a [BritishStandard7666])
+    sample.stringToScheme shouldBe a [BritishStandard7666]
   }
```
iRhysBradbury: Do not delete valid tests, then not replace the. its lazy.
Either, revert it back to an exception or add a test for this new bound.
Please do not modify codebase then remove tests for it. You _should fix_ the tests _not remove_ them.

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+  * @param uprn               uprn
+  * @param formattedAddress   cannonical address form
+  * @param paf                optional, information from Paf index
+  * @param nag                optional, information from Nag index
+  * @param geo                optional, geo information
+  * @param underlyingScore    score from elastic search
+  * @param underlyingMaxScore maximum score in the elastic result
+  */
+case class AddressResponseAddress(
+  uprn: String,
```
iRhysBradbury: formatting

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+  */
+case class AddressResponseError(
+  code: Int,
+  message: String
+)
+
+object AddressResponseError {
+  val emptyQuery = AddressResponseError(
+    code = 1,
+    message = "Empty query"
```
iRhysBradbury: messages

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+  )
+
+  val addressFormatNotSupported = AddressResponseError(
+    code = 2,
+    message = "Address format is not supported"
```
iRhysBradbury: messages

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+  )
+
+  val notFound = AddressResponseError(
+    code = 3,
```
iRhysBradbury: messages

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+import play.api.http.Status
+import play.api.libs.json.Json
+import uk.gov.ons.addressIndex.model.db.index.PostcodeAddressFileAddress
+
+/**
+  * Contains implicit readers/writers to convert response models into json (play standard).
+  * When needed, do following:
+  * `import uk.gov.ons.addressIndex.server.model.response.implicits._`
+  */
+object Implicits {
```
iRhysBradbury: Companions please

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+  )
+
+  val notFound = AddressResponseStatus(
+    code = Status.NOT_FOUND,
+    message = "Not Found"
+  )
+
+  val badRequest = AddressResponseStatus(
+    code = Status.BAD_REQUEST,
+    message = "Bad request"
```
iRhysBradbury: messages

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+
+object AddressResponseStatus {
+  val ok = AddressResponseStatus(
+    code = Status.OK,
+    message = "Ok"
+  )
+
+  val notFound = AddressResponseStatus(
+    code = Status.NOT_FOUND,
+    message = "Not Found"
```
iRhysBradbury: messages

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
```diff
@@ -0,0 +1,297 @@
...
+  */
+case class AddressResponseStatus(
+  code: Int,
+  message: String
+)
+
+object AddressResponseStatus {
+  val ok = AddressResponseStatus(
+    code = Status.OK,
+    message = "Ok"
```
iRhysBradbury: messages

model/src/main/scala/uk/gov/ons/addressIndex/model/AddressIndexModel.scala
Mironor: Well, it is still a scheme, but not yet supported.
Do you have a better name in mind?

model/src/test/scala/uk/gov/ons/addressIndex/model/AddressIndexModelTest.scala
Mironor: Codacy is complaining that one should not use the name of the class (AddressScheme) in the method

model/src/test/scala/uk/gov/ons/addressIndex/model/AddressIndexModelTest.scala
Mironor: Where are the removed tests? I just removed the exception as it made the code more complex without a need.
You don't throw an exception when you get no element in an option, you return `None`. Here we return `UnsupportedSchema`

model/src/main/scala/uk/gov/ons/addressIndex/model/AddressIndexModel.scala
Mironor: Well, it is still a scheme, but not yet supported.
Do you have a better name in mind?

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
Mironor: What about it?

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
Mironor: why messages?

iRhysBradbury: Just should be in messages although OK is pretty much lang independent

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
iRhysBradbury: empty lines, irregular comments, add these comments to scaladoc

server/app/uk/gov/ons/addressIndex/server/model/response/AddressResponse.scala
Mironor: This won't work (I tied), they should be imported in a specific order.

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
```diff
@@ -42,54 +43,105 @@ class AddressController @Inject()(esRepo : ElasticsearchRepository)(implicit ec
...
+      Ok(Json.toJson(AddressBySearchResponseContainer(
+        AddressResponse(
+          tokens,
+          addresses = addresses.map(AddressResponseAddress.fromPafAddress),
+          limit = 10,
+          offset = 0,
+          total = addresses.size
+        ),
+        AddressResponseStatus.ok,
+        errors = Seq()
```
iRhysBradbury: maybe default param?

Mironor: Should be shown explicitely, imo

iRhysBradbury: just seeing empty seq's everywhere. for future reference default params are explicitly applied not implicitly applied

Mironor:
```
case class A(errors: Seq[Any] = Seq.empty[Any])
val a = A()
```
Where is it implicitly applied?

iRhysBradbury: If you read my comments you'll see that I say they are **not implicitly applied**

Mironor: Sure, but it is not explicitly said in the object intitialisation.

iRhysBradbury: With named params, you can see that there are no errors :)

PR #27, opened 2016-11-30
Update application.conf to allow use of ENV labels

AlWestVT:
For deploying on Cloud Foundry we can use Environment Labels to specify the parameters used for elastic search with the API.

Tested in ONS Cloud Foundry

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -14,12 +14,17 @@ addressIndex {
   elasticSearch {
     local = false
     cluster = "06aee33092e6d4a954130175b435d54b"
+    cluster = ${?ONS_AI_API_ES_CLUSTER_NAME}
     // if you are using a remote server over SSL, try to use port 9343
     uri = "elasticsearch://06aee33092e6d4a954130175b435d54b.eu-west-1.aws.found.io:9343"
+    uri = ${?ONS_AI_API_ES_URI}
     shieldSsl = true
+    shieldSsl = ${?ONS_AI_API_SHIELD_SSL}
     shieldUser = "admin:changeMeImNotTheRealPassword"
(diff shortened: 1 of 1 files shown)


server/conf/application.conf
```diff
@@ -14,12 +14,17 @@ addressIndex {
   elasticSearch {
     local = false
     cluster = "06aee33092e6d4a954130175b435d54b"
+    cluster = ${?ONS_AI_API_ES_CLUSTER_NAME}
     // if you are using a remote server over SSL, try to use port 9343
     uri = "elasticsearch://06aee33092e6d4a954130175b435d54b.eu-west-1.aws.found.io:9343"
+    uri = ${?ONS_AI_API_ES_URI}
     shieldSsl = true
+    shieldSsl = ${?ONS_AI_API_SHIELD_SSL}
```
iRhysBradbury: Will this eval to a bool?

AlWestVT: Yes it resolves to a bool.  I actually negatively tested this aspect as I originally had a typo in the ENV label.  By default it is true so that is what was used in that instance.  When I corrected the ENV label it was set as false and shield SSL was disabled in the API.

server/conf/application.conf
```diff
@@ -14,12 +14,17 @@ addressIndex {
...
     shieldUser = "admin:changeMeImNotTheRealPassword"
+    shieldUser = ${?ONS_AI_API_SHIELD_USER}
```
analytically: can we name this shieldUsername?

analytically: or even shield.ssl, shield.username

iRhysBradbury: I don't mind changing, original name by @Mironor

PR #29, opened 2016-12-06
Implementing NAG format in the API

Mironor:


server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
```diff
@@ -88,6 +88,22 @@ class AddressController @Inject()(esRepo: ElasticsearchRepository)(implicit ec:
     }
   }
 
+  private def searchNagAddresses(tokens: AddressTokens): Future[Result] = {
+    esRepo.queryNagAddresses(tokens).map {
+      case NationalAddressGazetteerAddresses(addresses, maxScore) =>
+        Ok(Json.toJson(AddressBySearchResponseContainer(
```
iRhysBradbury: Improve this formatting

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -6,81 +6,71 @@ import uk.gov.ons.addressIndex.server.model.dao.ElasticClientProvider
...
-  def deleteAll() : Future[Seq[DeleteIndexResponse]] = {
-    deleteIndex(
-      PostcodeAddressFileAddress,
-      PostcodeIndex
-    )(client)
-  }
-
-  def queryUprn(uprn: String): Future[Option[PostcodeAddressFileAddress]] = client.execute{
+  def queryPafUprn(uprn: String): Future[Option[PostcodeAddressFileAddress]] = client.execute{
     search in pafIndex query { termQuery("uprn", uprn) }
```
iRhysBradbury: factor out repeat code

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -96,4 +86,21 @@ class AddressIndexRepository @Inject()(conf : AddressIndexConfigModule, elasticC
       )
     }
   }.map(response => PostcodeAddressFileAddresses(response.as[PostcodeAddressFileAddress], response.maxScore))
+
```
iRhysBradbury: factor out repeat code

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
```diff
@@ -75,11 +76,230 @@ class AddressResponseAddressSpec extends WordSpec with Matchers {
...
+      )
+
+      // When
+      val result = AddressResponseAddress.fromNagAddress(1)(givenNag)
+
+      // Then
+      result shouldBe expected
+    }
+
+    "be creatable from Elastic PAF response with wrong latitude" in {
```
iRhysBradbury: not sure these test names are describing the tests

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: This one has too many differences to make it factorable: index, query and the callback.
Factoring this code won't improve readability

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
Mironor: "Address response Address model should be creatable from Elastic PAF response with wrong altitude"
Maybe "Elastic PAF query" would be better, but suggest a better description if you have one in mind.

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
Mironor: How exactly? Put AddressBySearchResponseContainer on a new line?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: I tried it, but the factored code says that it `cannot find implicit for .as[T]` and just doesn't work. This is how I tried to do it:
```
  def queryPafUprn(uprn: String): Future[Option[PostcodeAddressFileAddress]] = queryUprn(uprn, pafIndex)

  def queryNagUprn(uprn: String): Future[Option[NationalAddressGazetteerAddress]] = queryUprn(uprn, nagIndex)

  private def queryUprn[T](uprn: String, index: String): Future[Option[T]] =  client.execute{
    search in index query { termQuery("uprn", uprn) }
  }.map(_.as[T].headOption)
```

If you have a better solution, I'm all for it

iRhysBradbury: You need to invoke with type params.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
iRhysBradbury: Repeat code is always bad.

expose buildingNumberFieldname and postcodeFieldName as arguments.

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
iRhysBradbury: Also, when using named params, name all the params please

```
 Ok(
  Json.toJson(
    AddressBySearchResponseContainer(
      AddressBySearchResponse(
        tokens,
         addresses = addresses.map(AddressResponseAddress.fromPafAddress(maxScore)),
         limit = 10,
         offset = 0,
         total = addresses.size
       ),
       OkAddressResponseStatus
     )
   )
)
```

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: Yes, I tried, it still doesn't compile

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
iRhysBradbury: "be creatable" Is not a start of a sentence, hence replace the test name.
Something more descriptive of whats going on here.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
iRhysBradbury: Type params must extend HitAs[T]

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: Writing unreadable code is worse then some repeated code (especially if it's only one repetition)
Exposing these arguments wont do any good, the query will change and will be very different for NAG and PAF so it would be better to put a query as a parameter. But there are also index name and a callback with different classes as well.

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
Mironor: This is all wrapped into `"Address response Address model" should {` so starting with "be creatable" does make sense, doesn't it.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: As saw together, this is not as trivial to fix as we thought. It's not cost effective to spend too much time on it and the fix may add more lines than it removes, so IMO this is good enough, I won't modify it any more.

iRhysBradbury: messy but ok

PR #31, opened 2016-12-09
Feature/token

AlWestVT:
Simple correction of EIGHT to EIGHTH in 3 Tokens


AlWestVT:
As this comes from the Python code @saniemi  should review this PR also.

iRhysBradbury:
Why is this "EIGHT"?
Reference: https://github.com/ONSdigital/address-index-data/blob/develop/DataScience/ProbabilisticParser/common/tokens.py#L68

iRhysBradbury:
This branch doesn't look quite right; I will fix; please don't merge.

saniemi:
Because I cannot spell.. this needs to be fixed in Python as well. I can make a hotfix, retrain a new model, and re-generate the validation results. I doubt that anything changes, but I think we should fix it  nonetheless, as it may change future behaviour. Happy to discuss.

PR #32, opened 2016-12-12
Adds following fields: classification, northing, easting to the API (by NAG)

Mironor:


RichardSmithONS:
I've just checked the code and it does indeed return latitude and longitude as well as easting and northing. API users will want to be able to get the coordinates in lat / lon for most services such as Google, but there are quite a few British OS-based services that use easting / northing instead. Converting on the client is hard work (e.g. reams of complex javascript).

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
```diff
@@ -85,220 +117,64 @@ class AddressResponseAddressSpec extends WordSpec with Matchers {
...
       // When
-      val result = AddressResponseAddress.fromNagAddress(1)(givenNag)
+      val result = AddressResponseAddress.fromNagAddress(1)(nag).geo
 
       // Then
-      result shouldBe expected
+      result shouldBe None
+    }
+
+    "be creatable from Elastic NAG response with invalid northing" in {
```
iRhysBradbury: If I've understood correctly;
"not be creatable"

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
```diff
@@ -2,10 +2,42 @@ package uk.gov.ons.addressIndex.server.model.response
 
 import org.scalatest.{Matchers, WordSpec}
 import uk.gov.ons.addressIndex.model.db.index.{NationalAddressGazetteerAddress, PostcodeAddressFileAddress}
-import uk.gov.ons.addressIndex.model.server.response.{AddressResponseAddress, AddressResponseGeo, AddressResponseNag, AddressResponsePaf}
+import uk.gov.ons.addressIndex.model.server.response.{AddressResponseAddress, AddressResponseGeo, AddressResponsePaf}
 
 class AddressResponseAddressSpec extends WordSpec with Matchers {
```
iRhysBradbury: a overall test description would be nice

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
```diff
@@ -185,7 +185,9 @@ object AddressResponseAddress {
     val geo: Try[AddressResponseGeo] = for {
```
iRhysBradbury: I think the naming of these classes should be revisited, consider namespacing instead of adding another prefix.

iRhysBradbury: eg:
```
object AddressIndex {
  object PostcodeAddressFile {
     case class Address()
     case class Response()
  }
 object NAG {
     case class Address()
     case class Response()
  }
}
```

Instead of names like `AddressPostcodeAddressFileAddress` (or something like that)

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
```diff
@@ -285,58 +293,83 @@ object AddressResponsePaf {
...
+case class AddressResponsePao(
+  paoText: String,
+  paoStartNumber: String,
+  paoStartSuffix: String,
+  paoEndNumber: String,
+  paoEndSuffix: String
+)
+
+object AddressResponsePao {
+  implicit lazy val addressResponsePaoFormat: OFormat[AddressResponsePao] = Json.format[AddressResponsePao]
```
iRhysBradbury: Should be a `Format`

server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec.scala
```diff
@@ -116,8 +116,12 @@ class ElasticsearchRepositorySpec extends WordSpec with SearchMatchers with Elas
         "locality" -> "n21",
         "organisation" -> "n22",
         "legalName" -> "n23",
-        "lat" -> "1.0000000",
-        "lon" -> "2.0000000"
+        "latitude" -> "1.0000000",
+        "longitude" -> "2.0000000",
+        "northing" -> "3",
+        "easting" -> "4",
+        "classificationCode" -> "n24"
```
iRhysBradbury: make these vars? repeat literals are dangerous

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
iRhysBradbury: Not PR Blocking

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
iRhysBradbury: PR Blocking

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
iRhysBradbury: Not PR Blocking

server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
iRhysBradbury: PR Blocking

server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec.scala
iRhysBradbury: PR Blocking

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
Mironor: They still need to be created in some external files. I did the prefix directly in the name to avoid naming collisions, but we can just use package prefix every time we are using them.

PR #35, opened 2016-12-15
Adds license

Mironor:


LICENSE.txt
```diff
@@ -0,0 +1,19 @@
+Copyright (C) 2015 Crown Copyright (Office of National Statistics)
```
saniemi: Should the year be 2016 or something else than 2015?

saniemi:
The main question I have is how this license relates to e.g. the CRFsuite package which is distributed under the modified BSD and libLBFGS which is distributed under the MIT license. Unfortunately, I cannot comment about the legal validity...

RichardSmithONS:
The default licence for ONS data and code is OGL3 http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/  Not sure what happens when third party software is incorporated.

PR #40, opened 2016-12-20
Fix how easting/northing is handled

Mironor:


server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
```diff
@@ -157,10 +157,10 @@ class AddressResponseAddressSpec extends WordSpec with Matchers {
...
         geo = Some(AddressResponseGeo(
-          latitude = 1.0d,
-          longitude = 2.0d,
-          easting = 3,
-          northing = 4
+          latitude = 50.7341677d,
+          longitude = -3.540302d,
+          easting = 291398,
+          northing = 93861
         )),
```
iRhysBradbury: separate out to common vars? literals inline are dangerous.

eg:

`val x = 6`

`x.toString`

Mironor: The string is "093861.00" for northing, where would you put the literal 93861? As a separate `val` in the spec?

PR #47, opened 2017-01-05
Modifies formatted address so that it contains commas

Mironor:


model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
```diff
@@ -173,10 +173,19 @@ object AddressResponseAddress {
 
     val poBoxNumber = if (paf.poBoxNumber.isEmpty) "" else s"PO BOX ${paf.poBoxNumber}"
 
+    val trimmedBuildingNumber = paf.buildingNumber.trim
+    val trimmedDependentThoroughfare = paf.dependentThoroughfare.trim
+    val trimmedThoroughfare = paf.thoroughfare.trim
+
+    val buildingNumberWithStreetName = if (trimmedDependentThoroughfare.isEmpty) {
```
iRhysBradbury: You could reduce the amount of code here with using an expression in your interpolated strings.

eg:
```
val buildingNumberWithStreetName = s"$trimmedBuildingNumber  ${if(!trimmedDependentThoroughfare.isEmpty) {s"$trimmedDependentThoroughfare, " }$trimmedThoroughfare"
```

iRhysBradbury: Im seeing some repeated 
`.map(_.trim).filter(_.nonEmpty).mkString(", ")`
Maybe make it a def?
`def delimit(strs: String*): String = strs.map(_.trim).filter(_.nonEmpty).mkString(", ")`
The impl looks prettier and is more descriptive of what you're doing here:
`delimit(paf.departmentName, paf.organizationName, paf.subBuildingName, paf.buildingName /* .....etc */)`

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/AddressResponse.scala
```diff
@@ -177,17 +177,17 @@ object AddressResponseAddress {
...
-    Seq(paf.departmentName, paf.organizationName, paf.subBuildingName, paf.buildingName,
+    val parts = Seq(paf.departmentName, paf.organizationName, paf.subBuildingName, paf.buildingName,
       poBoxNumber, buildingNumberWithStreetName, paf.doubleDependentLocality, paf.dependentLocality,
-      paf.postTown, paf.postcode).map(_.trim).filter(_.nonEmpty).mkString(", ")
+      paf.postTown, paf.postcode)
+
+    delimitByComma(parts)
   }
 
+  private def delimitByComma(parts: Seq[String]) = parts.map(_.trim).filter(_.nonEmpty).mkString(", ")
```
iRhysBradbury: Nice one, no need for `Seq` here, you can use varargs 

eg: 
`parts: String *`

It saves you making Seq's, and you can directly invoke

PR #51, opened 2017-01-09
Release/ai261

RichardSmithONS:
This change is to implement story 261 - paging parameters in the API, example URL http://localhost:9001/addresses?input=3%20EXETER%20EX2&format=bs&limit=10&offset=2

The limit and offset parameters are optional. Defaults and maxima are set in the application.conf.

Both parameters are checked for being numeric, less than the maximum and at least the minimum (0 for offset, 1 for limit). A Bad Request (400) is returned in each case with meaningful text.

The offset is incremented by 1 to become elastic4s's start value.

Offset, limit and size are returned in the response. If the offset is greater than the total number of results, it is not an error, rather size will be zero.

### DIFF
--- a/model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
+++ b/model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
@@ -5,7 +5,11 @@ case class ElasticSearchConfig(
   cluster: String,
   uri: String,
   indexes: IndexesConfig,
-  shield: ShieldConfig
+  shield: ShieldConfig,
+  defaultLimit: Int,
+  defaultOffset: Int,
+  maximumLimit: Int,
+  maximumOffset: Int
 )
 
 object ElasticSearchConfig {
@@ -21,7 +25,11 @@ object ElasticSearchConfig {
       ssl = true,
       user = "admin",
       password = ""
-    )
+    ),
+    defaultLimit=10,
+    defaultOffset=0,
+    maximumLimit=100,
+    maximumOffset=1000
   )
 }
 
--- a/server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
+++ b/server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
@@ -1,19 +1,25 @@
 package uk.gov.ons.addressIndex.server.controllers
 
 import javax.inject.{Inject, Singleton}
+
 import uk.gov.ons.addressIndex.server.modules.{AddressIndexActions, AddressParserModule, ElasticsearchRepository}
 import play.api.Logger
 import play.api.mvc.{Action, AnyContent}
+
 import scala.concurrent.ExecutionContext
 import com.sksamuel.elastic4s.ElasticDsl._
 import uk.gov.ons.addressIndex.crfscala.CrfScala.CrfTokenResult
+import uk.gov.ons.addressIndex.server.modules.AddressIndexConfigModule
 import uk.gov.ons.addressIndex.model.server.response._
 import uk.gov.ons.addressIndex.parsers.Implicits._
+
+import scala.util.Try
 
 @Singleton
 class AddressController @Inject()(
   override val esRepo: ElasticsearchRepository,
-  parser: AddressParserModule
+  parser: AddressParserModule,
+  conf : AddressIndexConfigModule
 )(implicit override val ec: ExecutionContext) extends AddressIndexController with AddressIndexActions {
 
   val logger = Logger("address-index-server:AddressController")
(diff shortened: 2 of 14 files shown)


model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
```diff
@@ -5,7 +5,11 @@ case class ElasticSearchConfig(
   cluster: String,
   uri: String,
   indexes: IndexesConfig,
-  shield: ShieldConfig
+  shield: ShieldConfig,
+  defaultLimit: Int,
+  defaultOffset: Int,
+  maximumLimit: Int,
```
iRhysBradbury: why is this in config?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexCannedResponse.scala
```diff
@@ -59,6 +59,54 @@ trait AddressIndexCannedResponse {
     )
   }
 
+  def LimitNotNumeric: AddressBySearchResponseContainer = {
```
iRhysBradbury: All of these follow the same pattern, to reduce the lines here do something like this:

```
private def BadRequestTemplate(errors: AddressResponseError*): AddressBySearchResponseContainer = {
     AddressBySearchResponseContainer(
       response = Error,
       status = BadRequestAddressResponseStatus,
       errors = errors
     )
 }
```

then you can reduce each of these to:

```
def OffsetNotNumeric: AddressBySearchResponseContainer = {
     BadRequestTemplate(OffsetNotNumericAddressResponseError)
}
```

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -99,9 +99,9 @@ class AddressIndexRepository @Inject()(
     }
   }
 
-  def queryPafAddresses(tokens: Seq[CrfTokenResult]): Future[PostcodeAddressFileAddresses] = {
+  def queryPafAddresses(start: Int, limit: Int, tokens: Seq[CrfTokenResult]): Future[PostcodeAddressFileAddresses] = {
```
iRhysBradbury: Have these as options and only construct the definintion if its `Some`

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
```diff
@@ -38,19 +44,45 @@ class AddressController @Inject()(
...
-      logger info s"#addressQuery parsed:\n${tokens.map(t => s"value: ${t.value} , label:${t.label}").mkString("\n")}"
-      formatQuery[AddressBySearchResponseContainer, Seq[CrfTokenResult]](
-        formatStr = format,
-        inputForPafFn = AddressQueryInput(tokens),
-        pafFn = pafSearch,
-        inputForNagFn = AddressQueryInput(tokens),
-        nagFn = nagSearch
-      ) getOrElse futureJsonBadRequest(UnsupportedFormat)
-    } getOrElse futureJsonBadRequest(EmptySearch)
+  def addressQuery(input: String, format: String, offset: Option[String] = None, limit: Option[String] = None): Action[AnyContent] = Action async { implicit req =>
```
iRhysBradbury: Why are these Strings?
(offset and limit)

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
iRhysBradbury: Eg:

MyType(message: String)
def(opt: Option[Int]): MyType = {
  opt.map { number => if(number > 10) { MyType("something1")} else {MyType("something2")} }.getOrElse(MyType("default"))
}

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
```diff
@@ -38,19 +44,45 @@ class AddressController @Inject()(
...
+    val maxOffset = conf.config.elasticSearch.maximumOffset
+    val limval = limit.getOrElse(defLimit.toString())
+    val offval = offset.getOrElse(defOffset.toString())
+    val limitInvalid = Try(limval.toInt).isFailure
+    val offsetInvalid = Try(offval.toInt).isFailure
+    val limitInt = Try(limval.toInt).toOption.getOrElse(defLimit)
+    val offsetInt = Try(offval.toInt).toOption.getOrElse(defOffset)
+// Check the offset and limit parameters before proceeding with the request
+    if (limitInvalid){
+      futureJsonBadRequest(LimitNotNumeric)
```
iRhysBradbury: not needed response, use play framework's built in responses

model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
RichardSmithONS: Having limits etc. externally configurable proved very useful for WDA (the ONS data API I still provide support for). An example is the maximum number of data cells allowed in the request which could be tuned during performance testing, and also set to different values in different environments. The Elasticsearch documentation talks of performance hits with "deep paging" so this may influence the maximum offset value allowed (although  I do concede that for address matching we're not usually interested in the lower results).

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
RichardSmithONS: This is to tidily deal with the case where the query string contains non-numeric values. For example limit=fifty. The  controller attempts to convert "fifty" to an Int and if this fails a bad request with a message string indicating that limit param must be numeric returned. If limit and offset were Option[Int]s I'm not sure how this scenario would  pan out.

iRhysBradbury: They shouldn't be. 
Checkout using `Option[Int]` with a default in the routes file.
Heres the Play Framework docs: 
https://www.playframework.com/documentation/2.5.x/ScalaRouting

more docs:
https://www.playframework.com/documentation/1.2.4/routes

Other refs:
http://mariussoutier.com/blog/2012/12/10/playframework-routes-part-1-basics/

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
iRhysBradbury: this wont be needed if the route param was a `Option[Int]`

RichardSmithONS: I'll give it a try. However, the downside with returning Play's built-in response (probably action not found - will see) rather than the APIs bespoke error response is that it is unfriendly to the client. In my experience API users like a predictable response format, so giving them different response types for different ways in which the parameter is wrong (e.g. limit=0 and limit=ten) might be unpopular.

model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
RichardSmithONS: _"It would be good to see the request only constructing the pagination (e4s) req if the option is some. It seems to be forced via defaults"._

Looking at implementing this one I see another issue. As I have coded it, the API's defaults are used if no parameters are specified. If no parameters are sent through to ES then its defaults will apply. I think it is preferable for the API to have control over the defaults.

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
iRhysBradbury: Whats the requirement here? To capture incorrect routes?

RichardSmithONS: Play does give a 400 using the built-in validation .....

Bad Request
For request 'GET /addresses?input=3%20EXETER%20EX2&format=bs&limit=10&offset=wibble' [Cannot parse parameter offset as Int: For input string: "wibble"]

but you don't get the nice formatted JSON response. I think it would be better to return the same response whatever is wrong with the parameter.
http://localhost:9001/addresses?input=3%20EXETER%20EX2&format=bs&limit=10&offset=-1

{
"response": {
"tokens": [],
"addresses": [],
"limit": 10,
"offset": 0,
"total": 0
},
"status": {
"code": 400,
"message": "Bad request"
},
"errors": [
{
"code": 7,
"message": "Offset parameter too small, minimum = 0"
}
]
}

iRhysBradbury: We interface using the scala client lib. We should aim to use the type system. this removes the need for all this.

RichardSmithONS:
I have implemented the bad request template. After discussing with Alan I'm staying with the API specifying the default page params, and the formatted json error message for non-numeric. I have marked the type-based validation changes as "todo" as I'd like to go through it F2F (so accept PR without for now, will apply later).

PR #53, opened 2017-01-11
Feature/bulk fix

iRhysBradbury:


iRhysBradbury:
fixing route priority clashes
changing route
adding content type
handling error and success
tested localy with public dev test api server
output accessible as JSON on console

iRhysBradbury:
@RichardSmithONS Currentlyy the file is uploaded and the response is logged out in the browser.
FYI this is a 1:M input:output

PR #55, opened 2017-01-13
Feature/crf tagger string

AlWestVT:
Adds building of Windows, Mac and Linux share object/DLL.  Utilises cross-compiling on Linux for Windows target.


alangorton:
LGTM

PR #65, opened 2017-01-26
Feature/hybrid index impl

iRhysBradbury:


iRhysBradbury:
Cody is complaining about a few things which are incorrect, eg, `.get` "on an option", then it's a `WSRequest` method.

iRhysBradbury:
@Ngwara @Mironor those conflicts above :(

PR #66, opened 2017-01-27
Feature/input pre processing

Mironor:
Adds pre-processing for the input (to Upper, remove counties, tokenize, replace synonyms)
Moves different collections of values to a separate files (asked by Tech Lead)
Replaces type aliases with simplier ones (asked by Tech Lead) in the parts of code that I worked on


parsers/src/main/resources/input_pre_post_processing/ordinal
```diff
@@ -0,0 +1,16 @@
...
+3RD
+FOURTH
+4TH
+FIFTH
+5TH
+SIXTH
+6TH
+SEVENTH
+7TH
+EIGHT
```
saniemi: My bad, the typo still exists. Shuold be "EIGHTH"

iRhysBradbury:
In what file can I view the preprocessing?

Mironor:
In `Tokens.scala` and `TokensTest.scala`

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -1,29 +1,67 @@
...
+    */
+  override def apply(input: String): Array[String] ={
+    val upperInput = input.toUpperCase()
+
+    val inputWithoutCounties = removeCounties(upperInput)
+
+    val tokens = inputWithoutCounties
+      .replace(" - ", "-")
+      .replace("- ", "-")
+      .replace(" -", "-")
```
saniemi: Sorry to be a pain, I know that I okay'ed the above, however, given todays work I think we need a slightly more complicated replace here. We want to remove the white spaces only if there are numbers around the hyphen, not if it's part of something else. In that case we actually want to remove it complete (my Python uses a regular expression to split the string before computing features).

I suggest we replace the above three replace lines with hyphens with something like this:
`\d+\s?-\s?\d+`
That matches with digits and optional white spaces. The replacement should be something like
`\d+-\d+`

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -1,29 +1,67 @@
...
+      .replace(",", " ")
+      .replace("\\", " ")
+      .split(" ")
+
+    replaceSynonyms(tokens).filter(_.nonEmpty)
+  }
+
+  private def removeCounties(input: String): String = {
+    val separatedCounties = county.mkString("|")
+    val countiesRegex = s"(?:\\b|\\s)($separatedCounties)(?:\\s|\\Z)"
```
iRhysBradbury: Are all `counties` Strings 'regex free'?

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -35,4544 +73,47 @@ object Tokens extends CrfTokenable {
...
+  lazy val outcodes: Seq[String] = fileToList(config.getString("parser.input-pre-post-processing.outcode"))
+
+  lazy val postTown: Seq[String] = fileToList(config.getString("parser.input-pre-post-processing.posttown"))
+
+  /**
+    * Contains key-value map of synonyms (replace key by value)
+    */
+  lazy val synonym: Map[String, String] = fileToList(config.getString("parser.input-pre-post-processing.synonym"))
+    // The "-1" in split is to catch the trailing empty space on the lines like "ENGLAND,"
+    .map(_.split(",", -1))
```
iRhysBradbury: is this safe? eg input string
`Rhys's Place, Somethingshire`

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
iRhysBradbury: I think this is what you might want to use - its catches space from the hyphen to user input
`\d+\s?.*?-.*?\s?\d+`

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
Mironor: They are, yes

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -1,29 +1,67 @@
...
+import com.typesafe.config.ConfigFactory
+import uk.gov.ons.addressIndex.crfscala.CrfScala.CrfTokenable
 
-//todo scaladoc this thing
+import scala.io.Source
+
+/**
+  * Tokenizer for the input
+  */
 object Tokens extends CrfTokenable {
```
iRhysBradbury: Should Tokens be a Class with instantiation params for the underlying file resources?
This would remove the usage of config in this class and move the concern of the config in the server from the lib impl

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
Mironor:
```
scala> "Rhys's Place, Somethingshire".split(",", -1)
res14: Array[String] = Array(Rhys's Place, " Somethingshire")
```
It's fine by me, or am I missing something?

iRhysBradbury: Is that your desired behaviour

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
iRhysBradbury: do you do any transformation/normalisation of the `separatedCounties` to normalise them?

eg:
```
def normalise(str: String): String = str.toLowerCase
val separatedCounties = Seq("Example").map(normalise)
val myInput = normalise("example")
```

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
Mironor: Yes and also the desired behavior is that `"ENGLAND,"` returns `Array("ENGLAND", "")`

iRhysBradbury: And this?: 
```
scala> "Rhys's Place, Somethingshire, test".split(",", -1)
res3: Array[String] = Array(Rhys's Place, " Somethingshire", " test")
``` 
:)

PR #68, opened 2017-01-30
Implements pagination logic

Mironor:
Credit to @RichardSmithONS , I just externilized his changes and moved them to a separate PR


RichardSmithONS:
Forgot to mention that the left page also needs the calculated min

PR #76, opened 2017-01-31
Feature/test coverage fix

iRhysBradbury:


server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec2.scala
```diff
@@ -0,0 +1,53 @@
+package uk.gov.ons.addressIndex.server.modules
```
Mironor: Those things are already tested, is there a need for additional ones?

iRhysBradbury: These tests are unit tests for the QueryDefinition objects.
The other tests are for the response from ES.

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
```diff
@@ -111,6 +118,12 @@ class AddressControllerSpec  extends PlaySpec with Results with AddressIndexCann
       Future.successful(HybridAddresses(Seq.empty, 1.0f, 0))
 
     override def client(): ElasticClient = ElasticClient.local(Settings.builder().build())
+
+    override def logger: Logger = ???
```
Mironor: I think adding additional public methods to the interface can be avoided.
The user of the interface don't need to know that it has logger, neither does he need to know that there are query generation methods as the user of the interface should never use them.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -16,27 +16,43 @@ import scala.concurrent.{ExecutionContext, Future}
...
+    * @param info
+    * @param t
+    * @param executable
+    * @tparam T
+    * @tparam R
+    * @tparam Q
+    * @return
     */
-  def queryAddresses(start: Int, limit: Int, tokens: Seq[CrfTokenResult]): Future[HybridAddresses]
+  def logExecute[T, R, Q](info: String)(t: T)(implicit executable: Executable[T, R, Q]): Future[Q] = {
```
Mironor: I saw this method in your PR, while it's neat, after a careful consideration, I would suggest to go without it as the code is more explicit (casting ES to DTO) this way.

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
iRhysBradbury: I'm not sure what you're asking; could you rephrase?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
iRhysBradbury: This method doesn't cast from ES to DTO.

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
Mironor: Those methods shouldn't be public as the user of the interface shouldn't use them.

server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec2.scala
iRhysBradbury: Good spot, the more tests the merrier?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: Yes, it also logs the query, but this functionality is already implemented directly in the method

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
iRhysBradbury: `logger` is used in `logExecute`
@AlWestVT mentioned he'd like to see more logs

server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec2.scala
Mironor: The one linked is complete (all tokens are provided), if you want to provide one where not all tokens are provided (I'd say that we can avoid doing it as query modification already requires us to modify corresponding 1 test), you may add them near the linked test in the same manner (the expected and the results are Json object instead of Strings)

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
Mironor: We already log queries
https://github.com/ONSdigital/address-index-api/blob/develop/server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala#L80
(and the same for query by uprn)

server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec2.scala
iRhysBradbury: I'm not sure what you're asking here - would you like me to remove these tests?

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
iRhysBradbury: I'm not sure what you're asking here - would you like me to remove these logs?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
iRhysBradbury: I'm not sure what you're asking here - would you like me to remove these logs?

server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec2.scala
Mironor: I'm saying that I already implemented those tests and there is no need in additional ones in my opinion.

server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
Mironor: I'm saying that the functionality is already implemented and there is no need in refactoring it now in my opinion.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: I'm saying that the functionality is already implemented and there is no need in refactoring it now in my opinion.

PR #79, opened 2017-02-05
Bulk requests (baseline) backend

Mironor:
Implements the backend logic for bulk requests (without using bulk ES feature).

`Await.result` is intentional (we should wait for previous batch before starting a new one)


parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/AddressParser.scala
```diff
@@ -12,6 +12,10 @@ import uk.gov.ons.addressIndex.crfscala.{CrfAggregateFeature, CrfFeature, CrfFea
   */
 object AddressParser extends CrfParser {
 
+  val currentDirectory = new java.io.File(".").getCanonicalPath
+  val modelPath = s"$currentDirectory/parsers/src/main/resources/addressCRFA.crfsuite"
```
iRhysBradbury: This value is moved to config eg: https://github.com/ONSdigital/address-index-api/pull/77/files

iRhysBradbury:
Peer with @Ngwara and merge squash was made

PR #82, opened 2017-02-07
Feature/json bulk api

iRhysBradbury:


server/test/uk/gov/ons/addressIndex/server/controllers/AddressControllerSpec.scala
```diff
@@ -352,7 +354,36 @@ class AddressControllerSpec  extends PlaySpec with Results with AddressIndexCann
       status(result) mustBe BAD_REQUEST
       actual mustBe expected
     }
-    
+
+    "reply on bulk post req" ignore new WithApplication {
```
Mironor: If anyone sees this, why is this test ignored?

iRhysBradbury: Good spot

PR #90, opened 2017-02-08
adding postcode validation and token pre-processing. adding supportin…

iRhysBradbury:
…g tests

### DIFF
--- a/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
+++ b/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
@@ -55,7 +55,7 @@ object Tokens extends CrfTokenable {
 
   override def normalise(tokens: Array[String]): Array[String] = tokens.map(_.toUpperCase)
 
-  def tokensToMap(tokens: Seq[CrfTokenResult]): Map[String, String] = {
+  def tokensToValidMap(tokens: Seq[CrfTokenResult]): Map[String, String] = {
     val validatedTokens = validatePostCode(tokens.groupBy(_.label))
     validatedTokens.map {
       case (token, values) => (token, values.map(_.value).mkString(" "))
--- a/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
+++ b/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
@@ -82,7 +82,7 @@ object Tokens extends CrfTokenable {
             )
           )
         } else {
-          tokens.-(Tokens.postcode)
+          tokens - Tokens.postcode
         }
       } else {
         tokens
--- a/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
+++ b/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
@@ -66,7 +66,7 @@ object Tokens extends CrfTokenable {
     val postCodeTokens = tokens.getOrElse(Tokens.postcode, Seq.empty)
     if(postCodeTokens.size == 1) {
       val token = postCodeTokens.head.value
-      if(token.length >= 5) {
+      if(token.length >= 4) {
         val outcode = token.substring(token.length - 3, token.length)
         val incode = token.substring(0, token.indexOf(outcode))
         val regex = "[0-9][A-Z][A-Z]".r
--- a/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
+++ b/parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
@@ -4,6 +4,7 @@ import com.typesafe.config.ConfigFactory
 import uk.gov.ons.addressIndex.crfscala.CrfScala.{CrfTokenResult, CrfTokenable}
 
 import scala.io.Source
+import scala.util.matching.Regex
 
 /**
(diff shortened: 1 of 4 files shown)


parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -55,11 +56,43 @@ object Tokens extends CrfTokenable {
...
   }
 
+  def validatePostCode(tokens: Map[String, Seq[CrfTokenResult]]): Map[String, Seq[CrfTokenResult]] = {
+    val postCodeTokens = tokens.getOrElse(Tokens.postcode, Seq.empty)
+    if(postCodeTokens.size == 1) {
+      val token = postCodeTokens.head.value
+      if(token.length >= 5) {
+        val outcode = token.substring(token.length - 3, token.length)
+        val incode = token.substring(0, token.indexOf(outcode))
+        val regex = "[0-9][A-Z][A-Z]".r
```
saniemi: In the future we may wish to consider using
` [0-9][A-Z-[CIKMOV]]{2}`
or something more robust, but for time being this is fine.

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -55,11 +56,43 @@ object Tokens extends CrfTokenable {
...
+    validatedTokens.map {
       case (token, values) => (token, values.map(_.value).mkString(" "))
     }
   }
 
+  def validatePostCode(tokens: Map[String, Seq[CrfTokenResult]]): Map[String, Seq[CrfTokenResult]] = {
+    val postCodeTokens = tokens.getOrElse(Tokens.postcode, Seq.empty)
+    if(postCodeTokens.size == 1) {
+      val token = postCodeTokens.head.value
+      if(token.length >= 5) {
```
saniemi: Why 5? I thought that it should be >= 4?
"X1AA" -> "X 1AA"?

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -55,11 +56,43 @@ object Tokens extends CrfTokenable {
...
+            key = Tokens.postcode,
+            value = Seq(
+              CrfTokenResult(
+                value = s"$incode $outcode",
+                label = Tokens.postcode
+              )
+            )
+          )
+        } else {
+          tokens - Tokens.postcode
```
saniemi: My Scala is non existing, so apologies if this is irrelevant. I assume that the - operator here removes the postcode token from the sequence of tokens, correct? If so, then it's correct.

iRhysBradbury: Correct :)

ref to func : http://www.scala-lang.org/api/2.12.x/scala/collection/immutable/Map.html#-(key:K):scala.collection.Map[K,V]

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
iRhysBradbury: Good spot!

parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -55,11 +56,43 @@ object Tokens extends CrfTokenable {
   override def normalise(tokens: Array[String]): Array[String] = tokens.map(_.toUpperCase)
 
   def tokensToMap(tokens: Seq[CrfTokenResult]): Map[String, String] = {
-    tokens.groupBy(_.label).map {
+    val validatedTokens = validatePostCode(tokens.groupBy(_.label))
```
Mironor: The aim of the function `tokensToMap` is to transform a list of tokens into a `Map` with a label as a key and concatenated value as a value, why do we have a validation (of the postcode) inside of it?

iRhysBradbury: Hi alex, we need to do this to all postcode tokens, all the time

Mironor: Then it should be in the `apply` method, shouldn't it?

iRhysBradbury: It's a little different to that, this is transforming the tokens to valid map of tokens to be used in the query, changing func name

parsers/src/test/scala/uk/gov/ons/addressIndex/parsers/PostCodeValidationTest.scala
```diff
@@ -0,0 +1,157 @@
...
+  it should "X1AA -> X 1AA" in {
+    testHelper(
+      input = "X1AA",
+      expected = "X 1AA"
+    )
+  }
+
+  //These ingored tests are bounds tests for postcode, these occurrences will never happen
+  //because the parser will pick them up as building number.
+  //still interesting to keep.
```
Mironor: If you want to keep ignored tests as a documentation, please add the information that they contain to the javadoc instead and remove the ignored tests

iRhysBradbury: This is a special request from @Ngwara

PR #101, opened 2017-02-13
Adds parameters to the config file so that we could modify some of the query parameters

Mironor:


server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -126,7 +124,7 @@ class AddressIndexRepository @Inject()(
             matchQuery(
               field = "lpi.paoStartNumber",
               value = token
-            )),
+            ).boost(queryParams.buildingNumberLpiBoost)),
           tokensMap.get(Tokens.locality).map(token =>
```
saniemi: Do we test Tokens.locality against PAF.DEPENDENT_LOCALITY? That should be included if not...

Mironor: Nope, not currently the case, do you want a configurable boost on it ?

Mironor:
@RichardSmithONS we did set all the default parameters in the config.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
saniemi: I don't think we need a configurable boost on Locality at this point. We need to modify the query anyway after we have re-indexed.

PR #102, opened 2017-02-14
Feature/clericalpart2

RichardSmithONS:
These are more UI changes for clerical tool - 311 is the redesigned tokenisation output (nag output fields may require tweaking) and 260 logical status with lookup. I have also included some minor cosmetic tweaks requested by Taiwo.


Mironor:
May I ask why was this merged to DEV without travis going through it? It doesn't look like a trivial change to me and it also does not compile:
https://travis-ci.org/ONSdigital/address-index-api/builds/201537289

PR #108, opened 2017-02-14
Feature/post tokenizer treatement

Mironor:


parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -27,9 +56,8 @@ object Tokens extends CrfTokenable {
     val inputWithoutCounties = removeCounties(upperInput)
 
     val tokens = inputWithoutCounties
-      .replace(" - ", "-")
-      .replace("- ", "-")
-      .replace(" -", "-")
+      .replaceAll("(\\d+) ?- ?(\\d+)", "$1-$2")
```
iRhysBradbury: Possible improvement to this regex

`.replaceAll("(\\d+).*?-.*?(\\d+)", "$1-$2")`

parsers/src/test/scala/uk/gov/ons/addressIndex/parsers/postTokenizeTreatmentTest.scala
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+    // When
+    val actual = Tokens.postTokenizeTreatment(input)
+
+    // Then
+    actual shouldBe expected
+  }
+
+  it should "NOT transform buildingName if it doesn't follow the pattern" in {
```
iRhysBradbury: should we add another test for input of 

"40  -      45"

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -113,114 +113,127 @@ class AddressIndexRepository @Inject()(
     * @param tokens tokens for the ES query
     * @return Search definition containing query to the ES
     */
-  def generateQueryAddressRequest(tokens: Seq[CrfTokenResult]): SearchDefinition = {
-
-    val tokensMap = Tokens.tokensToValidMap(tokens)
+  def generateQueryAddressRequest(tokens: Map[String, String]): SearchDefinition = {
 
     val query =
       bool {
```
iRhysBradbury: should we look to factor out all this repeat code into one manageable block?

parsers/src/test/scala/uk/gov/ons/addressIndex/parsers/postTokenizeTreatmentTest.scala
Mironor: At this stage the tokens is "40-45" and that is already covered by the 120-122 case.
If the token is "40 - 45" it's a problem with pre-treatment, not with the post-treatment

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: It is fine like it currently is (explicit).
It is soon to be rewritten completely, so it's pointless to try to refactor it currently.

PR #109, opened 2017-02-15
replace addressCRFA.crfsuite

RichardSmithONS:
This is a simple replace of the model file addressCRFA.crfsuite with a new one provided by Sami. 
I have tested that the API still works locally (returns something sensible) and the unit tests pass.


iRhysBradbury:
Comments on model version please

saniemi:
Richard, sounds good. Have you run the verification using 100k addresses?

saniemi:
The parser tests I use are here:
https://github.com/ONSdigital/address-index-data/blob/develop/DataScience/ProbabilisticParser/tests/test_parser.py
You may wish to consider implementing something similar.

iRhysBradbury:
Good to note why this model has been replaced; supporting more welsh chars and more org names

saniemi:
In short, a new re-trained model version was produced to:
- Improve parsing of addresses written in Welsh
- Improve parsing of input strings with missing address components
- Improve parsing of strings that may contain only a single label such as organisation or street name

Parameter optimisation to produce this version of the model was carried out using the full sequence precision rather than the individual token precision.

PR #110, opened 2017-02-15
Feature/no found matches

iRhysBradbury:


demo-ui/app/uk/gov/ons/addressIndex/demoui/views/MatchTypeHelper.scala
```diff
@@ -4,7 +4,7 @@ object MatchTypeHelper {
 
   def matchType(addressId: String, ids: Seq[String], matchedFormattedAddress: String): String = {
     val matchedIds = ids.count(_ == addressId)
-    if(matchedIds == 1) {
+    if(matchedIds == 1 && !matchedFormattedAddress.isEmpty) {
```
Mironor: `!matchedFormattedAddress.isEmpty` is equivalent to `matchedFormattedAddress.nonEmpty`

iRhysBradbury: Thanks

Mironor:
I wonder, if the problem is only with bulk search, why add "default" empty address on empty result to the repository instead of adding this `if` to the `bulkQuery` method ( here https://github.com/ONSdigital/address-index-api/blob/develop/server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala#L213 )? 
This way we localize the change to the relevant functionality and avoid any regressions in other parts of the application.

iRhysBradbury:
@Mironor agree - changing now

PR #112, opened 2017-02-15
Feature/search backpressure

Mironor:


model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
```diff
@@ -23,7 +22,11 @@ case class ShieldConfig(
 case class AddressIndexConfig(
   parserLibPath: String,
   bulkLimit: Int,
-  elasticSearch: ElasticSearchConfig
+  elasticSearch: ElasticSearchConfig,
+  bulkRequestsPerBatch: Int,
+  bulkMiniBatchUpscale: Float,
+  bulkMiniBatchDownscale: Float,
```
iRhysBradbury: Proposed structure:
```
bulk {
   miniBatch {
     upscale
     //etc
  }
}
``

**Comment by iRhysBradbury** [comment]
codacy complaining about trivial things

File: server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala [hunk 2]
```diff
@@ -128,68 +130,85 @@ class AddressController @Inject()(
...
+    else logger.info(s"#bulkQuery sending a mini-batch of the size $miniBatchSize")
+
+    val miniBatch = requests.take(miniBatchSize)
+    val requestsAfterMiniBatch = requests.drop(miniBatchSize)
+    val result: BulkAddresses = Await.result(queryBulkAddresses(miniBatch, conf.config.bulkLimit), Duration.Inf)
+
+    val requestsLeft = result.failedRequests ++ requestsAfterMiniBatch
+
+    if (requestsLeft.isEmpty) successfulResults ++ result.successfulBulkAddresses
+    else if (miniBatchSize == 1 && result.failedRequests.nonEmpty) throw new Exception("Bulk query request: mini-bulk was scaled down to the size of 1 and it still fails, something's wrong with ES.")
```
iRhysBradbury: Why not throw the exception from ES?

iRhysBradbury:
Will this horizontally scale?

Mironor:
It is not, but there is no need to scale it horizontally (in my opinion) because this code is not the bottleneck. The bottleneck is currently the ES. To scale it horizontally we would need to use akka or (easier) Spark.

server/app/uk/gov/ons/addressIndex/server/controllers/AddressController.scala
Mironor: The exception from ES is already logged (as an `INFO`) and I would need to add a field to `BulkAddressRequestData` to store the "bubbled" exception message just to show it in this warning. It may be worth it, let me meditate on it.

AlWestVT: Cloud Foundry scales horizontally so all code should be in line with the 12-Factor application guidelines.

AlWestVT:
Cloud Foundry scales horizontally so all code should be in line with the 12-Factor Application Design guidelines.

iRhysBradbury:
should we have some representative tests of what this is addressing? (5k req's)

PR #116, opened 2017-02-17
filtering by 66%

iRhysBradbury:


iRhysBradbury:
Some info around this logic.
Currently in the bulk ui, we return sets of 5.
The percentage shown is a percentage within the set of 5 eg:
score = (document score/max(score in set of 5)) * 100.
This is still being shown. 
Under the hood of the UI we do some filtering.
We calculate another column under the hood which is:
under the hood score =(document score/max(score of the full results across sets)) * 100.
We then take `under the hood score` and filter it if it's greater than or equal to 66%.
This is to remove relatively bad match results.

PR #125, opened 2017-02-24
Adapts API to the new index (without modifying the query)

Mironor:


server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -106,7 +107,7 @@ class AddressIndexRepository @Inject()(
           tokens.get(Tokens.buildingNumber).map(token =>
             matchQuery(
               field = "lpi.paoStartNumber",
-              value = token
+              value = Try(token.toShort).getOrElse(null)
```
Mironor: This is a small modification of the old query just to make the tests pass, this will be overwritten with the new query

PR #126, opened 2017-02-24
Feature/ai407

RichardSmithONS:
Implementation of the new query.

Extensive changes to the AddressIndexRepository, ElasticSearchRepositorySpec and Configuration (new hierarchical query settings objects). It is intended that the boost settings are externalised but not fuzziness for now. 

Note that the query has a much more complex structure, but it has been tested against the old index. Some extra query parts for the split postcode and _PAF and _NAG yet to be added.


saniemi:
Open points e.g. fuzziness, pafAll, and MinimumShuoldMatch were discussed. Happy to review when changes have been made.

Mironor:
@saniemi Changes have been made

model/src/main/scala/uk/gov/ons/addressIndex/model/config/AddressIndexConfig.scala
```diff
@@ -12,13 +12,29 @@ case class ElasticSearchConfig(
...
+  streetName: StreetNameConfig,
+  townName: TownNameConfig,
+  postcode: PostcodeConfig,
+  organisationName: OrganisationNameConfig,
+  departmentName: DepartmentNameConfig,
+  locality: LocalityConfig,
+  pafAllBoost: Float,
+  nagAllBoost: Float,
+  defaultBoost: Float,
+  minimumShouldMatch: String
```
saniemi: I reckon minimumShouldMatch can be dropped as we are no longer using it.

RichardSmithONS:
I have followed through the postocde in and out and it all  looks fine with thorough tests. New parts of query should work when new index is loaded.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -96,142 +100,306 @@ class AddressIndexRepository @Inject()(
...
+          matchQuery(
+            field = "paf.buildingName",
+            value = token
+          ).boost(queryParams.buildingName.pafBuildingNameBoost)),
+        tokens.get(Tokens.buildingName).map(token =>
+          matchQuery(
+            field = "lpi.paoText",
+            value = token
+          ).boost(queryParams.buildingName.lpiPaoTextBoost))
+      ).flatten.map(_.fuzziness("AUTO"))
```
saniemi: Do we know what "AUTO" does? Does it use scaling i.e. if len(string) < x then fuzziness = 1, else 2?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -96,142 +100,306 @@ class AddressIndexRepository @Inject()(
...
+          bool(must(saoQuery).should(subBuildingNameQuery)),
+          bool(must(paoQuery).should(buildingNameQuery)),
+          bool(should(buildingNumberQuery)),
+          bool(should(streetNameQuery)),
+          bool(should(townNameQuery)),
+          bool(must(postcodeQuery).should(postcodeInOutQuery))
+        )).should(Seq(
+          bool(should(organisationNameQuery)),
+          bool(should(departmentNameQuery)),
+          bool(should(localityQuery)),
```
saniemi: Looks really good! In the future we may wish to consider changing the localityQuery to be more like townNameQuery because of the lack of definition for towns and localities. However, for time being, we should use this solution.

server/conf/application.conf
```diff
@@ -33,51 +33,116 @@ addressIndex {
...
+        pafWelshDoubleDependentLocalityBoost=${?ONS_AI_API_QUERY_LOCALITY_PAF_WELSH_DOUBLE_DEPENDENT_LOCALITY_BOOST}
+      }
+      pafAllBoost=0.5
+      pafAllBoost=${?ONS_AI_API_QUERY_PAF_ALL_BOOST}
+      nagAllBoost=0.5
+      nagAllBoost=${?ONS_AI_API_QUERY_NAG_ALL_BOOST}
+      defaultBoost=1.0
+      defaultBoost=${?ONS_AI_API_QUERY_DEFAULT_BOOST}
       minimumShouldMatch="45%"
       minimumShouldMatch=${?ONS_AI_API_QUERY_MINIMUM_SHOULD_MATCH}
```
saniemi: I reckon we can remove the minimumShouldMatch as it's not used?

server/conf/application.conf
```diff
@@ -33,51 +33,116 @@ addressIndex {
...
     }
```
saniemi: Can we also change line 156 perBatch to 350 or something slightly higher than 100, or are we going to do it in a separate PR after reconfigured ES is online?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: https://www.elastic.co/guide/en/elasticsearch/guide/current/fuzziness.html
```
The fuzziness parameter can be set to AUTO, which results in the following maximum edit distances:

0 for strings of one or two characters
1 for strings of three, four, or five characters
2 for strings of more than five characters
Of course, you may find that an edit distance of 2 is still overkill, and returns results that don’t appear to be related. You may get better results, and better performance, with a maximum fuzziness of 1.
```

saniemi: Thank you. This is the sought behaviour.

server/conf/application.conf
Mironor: We don't know the limit with the new query yet. We'll update it after the first tests

saniemi: Fair point.

PR #127, opened 2017-02-26
Adds splunk logs to the controller

Mironor:


Mironor:
Number of hits is currently irrelevant (in my opinion), we get hundreds of thousands of results and most of them are irrelevant.
Highest match score is not useful with current Scorer, when the new one is implemented, we will decide if it needs to be in splunk.

RichardSmithONS:
Number of hits may be of analytical interest for zero results (which is a 200 response). Presumably if there are zero results the highest score would be zero so this metric could be used to determine how many requests have not matched any addresses. Agreed a _meaningful_ number (the Hopper Matchscore?) would be of much more values than stats based on the raw elastic score.

PR #128, opened 2017-02-28
Feature/ai311

RichardSmithONS:
Changes to complete tokenised address changes for Clerical tool in sprint 9.

Postcode moved to separate section to avoid ambiguity

"rowspan" done with CSS

postcode input tokens anticipated - not sure if split postcode will be part of output.


demo-ui/conf/messages.en
```diff
@@ -10,6 +10,7 @@ clerical.ta=Tokenised Address
 clerical.token_description=Token name
 clerical.fieldnamePAF=Field Name (PAF)
 clerical.fieldnameNAG=Field Name (NAG)
+clerical.fieldnamederived=Field Name (PAF or NAG)
```
Mironor: There should underlines between words to make it more readable

PR #138, opened 2017-03-06
Copies paoStartSuffix to building name queries as it is a building name

Mironor:


saniemi:
This is only true if the BuildingName is of form "number+suffix" e.g. BuildingName=65B -> PAO_START_NUMBER = 65, PAO_START_SUFFIX=B. Not applicable if BuildingName = "Alex Lodge".

PR #140, opened 2017-03-07
Uses constant scoring instead of tf/idf

Mironor:


RichardSmithONS:
I'm not clear on what effect this actually has - ES docco quite terse: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-constant-score-query.html

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -105,303 +105,305 @@ class AddressIndexRepository @Inject()(
...
-        matchQuery(
+        constantScoreQuery(matchQuery(
           field = "lpi.saoText",
           value = token
-        ).boost(queryParams.subBuildingName.lpiSaoTextBoost)),
+        )).boost(queryParams.subBuildingName.lpiSaoTextBoost)),
       tokens.get(Tokens.saoStartSuffix).map(token =>
-        matchQuery(
+        constantScoreQuery(matchQuery(
           field = "lpi.saoStartSuffix",
```
saniemi: I am not sure I understand this here. if we have subBuildingName='FLAT 1', then SAO_START_NUMBER should be 1 without suffix and covered by saoQuery. If we have subBuildingNAme='120A-122B', then we need SAO_START_NUMBER=120, SAO_START_SUFFIX=A, SAO_END_NUMBER=122, and SAO_END_SUFFIX=B, which again should be in the saoQuery and not needed here. Have I misunderstood something? Maybe these two need to be joined?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -105,303 +105,305 @@ class AddressIndexRepository @Inject()(
...
       paoQuery,
-      saoQuery,
+      saoQuery
+      // `dismax` dsl does not exist, `: _*` means that we provide a list (`queries`) as arguments (args) for the function
+    ).filter(_.nonEmpty).map(queries => dismax.query(queries: _*).tieBreaker(queryParams.disMaxTieBreaker))
+
+    val preciseShouldQuery = Seq(
       organisationNameQuery,
       departmentNameQuery,
       localityQuery
```
saniemi: Fine for now, but after results are known, we should test if changing this to similar as the townNameQuery and moving it inside the preciseMustQuery helps to limit some unwanted matches.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -105,303 +105,305 @@ class AddressIndexRepository @Inject()(
...
-        matchQuery(
+        constantScoreQuery(matchQuery(
           field = "lpi.paoText",
           value = token
-        ).boost(queryParams.buildingName.lpiPaoTextBoost)),
+        ).fuzziness(defaultFuzziness)).boost(queryParams.buildingName.lpiPaoTextBoost)),
       tokens.get(Tokens.paoStartSuffix).map(token =>
-        matchQuery(
+        constantScoreQuery(matchQuery(
           field = "lpi.paoStartSuffix",
```
saniemi: Same comments as in case of subBuildingName. BuildingName could be '66A-68D' and then PAO_START_NUMBER=66, PAO_START_SUFFIX=A, PAO_END_NUMBER=68, and PAO_END_SUFFIX=D. Perhaps we need to revisit this and the above when we add the parser post processing to extract the numbers and suffixes from these type of ranges.

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
Mironor: We don't currently process subBuildingName (as we do with Building name)

Mironor: You mean sao query should not be dis_max'ed?

saniemi: No, I reckon it should. I think we are find for now, but in the long term we need to think how we may wish to group subBuildingName and BuildingName together, if they are joined using "must".

PR #141, opened 2017-03-07
Feature/custodian

RichardSmithONS:
This is all three custodian tickets - 282, 197 and 329.

For 282 a new lookup is added to the model. This fetches the LA Name and ONS Geography code.

197 and 329 enable the UI to display the values via Clerical Tool or single match (n.b. not showing ONS geog code at the mo but could).


RichardSmithONS:
I  think Codacy's complaint can be ignored, it's simply that the NationalAddressGazetteerAddress object is quite big.

RichardSmithONS:
After discussion with Sarah and Ibukun we have decided to move the custodian lookup to a file as a separate story rather than do it immediately.

PR #143, opened 2017-03-08
add fetch by uprn to demo ui (446)

RichardSmithONS:
This PR includes changes to make the single search interpret a number-only input as a UPRN. The Clerical tool still attempts a search.

I also changed the code displayed in the local custodian to the ONS one, and removed the redundant format from the URLs and views.


RichardSmithONS:
Forgot to mention the URI of the external API has been updated in the application.conf to the new IBM one.

demo-ui/app/uk/gov/ons/addressIndex/demoui/controllers/SingleMatchController.scala
```diff
@@ -152,6 +161,65 @@ class SingleMatchController @Inject()(
       }
     }
   }
+
+  /**
+    * Perform match by calling API with address string. Can be called directly via get or redirect from form
+    * @param input
+    * @return result to view
+    */
+  def doGetUprn(input : String, page: Option[Int]) : Action[AnyContent] = Action.async { implicit request =>
```
Mironor: I don't think we need page/limit/offset for getByUprn method as it will return either one or no elements.

demo-ui/app/uk/gov/ons/addressIndex/demoui/controllers/SingleMatchController.scala
```diff
@@ -152,6 +161,65 @@ class SingleMatchController @Inject()(
...
+        pageSize = pageSize,
+        pageMax = maxPages,
+        addressBySearchResponse = None,
+        classification = None)
+      Future.successful(
+        Ok(viewToRender)
+      )
+    } else {
+      logger info("Single Match with supplied input address " + addressText)
+      val testuprn = BigInt(addressText)
```
Mironor: Could we have a better naming for this value?

PR #146, opened 2017-03-09
Add relatives to API output

RichardSmithONS:
After discussion, we agreed that for JIRA 283 we would simply add the relatives object to the API response unconditionally.

Note that I also tidied the routes file a little and left in a small tryout CSS change on the UI for the footer.


server/test/uk/gov/ons/addressIndex/server/modules/ElasticsearchRepositorySpec.scala
```diff
@@ -70,13 +72,16 @@ class ElasticsearchRepositorySpec extends WordSpec with SearchMatchers with Elas
...
+  val hybridNagRelatives = "[]"
   // Fields with this value are not used in the search and are, thus, irrelevant
   val hybridNotUsed = ""
   val hybridNotUsedNull = null
 
   // Secondary PAF/NAG is used for single search (to have some "concurrence" for the main address)
   // and in the Multi Search
   val hybridSecondaryUprn = 2L
+  val hybridSecondaryParentUprn = 4L
+  val hybridSecondaryRelatives ="h03r"
```
Mironor: Relatives is actually an `Array[Map[String, Any]]`. As an example:
https://github.com/ONSdigital/address-index-data/blob/develop/batch/src/test/scala/uk/gov/ons/addressindex/utils/SqlHelperSpec.scala#L152

model/src/main/scala/uk/gov/ons/addressIndex/model/db/index/HybridAddress.scala
```diff
@@ -63,6 +67,8 @@ object HybridAddress {
 
       HybridAddress(
         uprn = hit.sourceAsMap("uprn").toString,
+        parentUprn = hit.sourceAsMap("parentUprn").toString,
+        relatives = hit.sourceAsMap("relatives").toString,
```
Mironor: This won't work with `relatives` as it is an `Array[Map[String, Any]]`
Look at how the lpi and paf fields are analysed and it should be more or less the same for `relatives`

RichardSmithONS:
Now ready for merge

PR #165, opened 2017-03-27
Add results to the query-debug page

Mironor:


Mironor:
If the result of the method is wrapped into `Future`, you will need to use `Action.async`. There is no need to use `Action.async` otherwise as even the plain `Action` request is already asynchronous under the hood.
Let's not use infix format (the `log info "log data"` one), it's confusing and unnecessary.

PR #167, opened 2017-03-27
Update application.conf

saniemi:
Changing tie breaker default from 0 to 0.5. This helps with addresses with number and suffix ranges. The SAO test dataset went from 97 to 13 false positives.


Mironor:
Does it work good on the edge cases?

saniemi:
Good question - yes, it does work on EdgeCases. It reduces the number of false positives by 7, which isn't as much as one would hope, but at least it improves things. I will be running other datasets soon.

PR #168, opened 2017-03-28
Fixes the problem with the post-tokenize-treatment order

Mironor:


parsers/src/test/scala/uk/gov/ons/addressIndex/parsers/postTokenizeTreatmentTest.scala
```diff
@@ -21,19 +18,16 @@ class postTokenizeTreatmentTest extends FlatSpec with Matchers {
     )
 
     // When
-    val actual = Tokens.postTokenizeTreatment(input)
+    val actual = Tokens.postTokenizeTreatmentBuildingName(input)
```
saniemi: You are now testing only the  BuildingName treatment, not the complete treatment. Is there a possibility that any of the other methods could change the tokens or their content leading to unwanted results?

Mironor: Nice question, thanks for asking.
If I leave the `postTokenizeTreatment` as before, it will add `FLAT A` to the subBuildingName token. Now we have other method influencing the result. I think, instead of addind `FLAT A` to the expected result, we'd better reduce the granularity of the code that is tested by this method (make it more Atomic).

saniemi: Sounds good. Approved.

PR #194, opened 2017-04-12
Version in banner and Navbar changes for Demo UI

RichardSmithONS:
Three JIRAs  (468, 454, 474) - Version number and Epoch obtained as lazy values from a singleton version module similar to the one in server project.  A shared banner viewlet then fetches the values, extracts the crucial 6 (7?) characters of the git commit id and makes them into a non-underlined link as Taiwo suggested.

The navbar changes are to turn it into a dropdown on a small screen (works fine on my phone, and you can see on a desktop as you narrow the window)  and add the navbar to the error view.


RichardSmithONS:
Forgot to mention that I tried a different image for the debug tool - Taiwo is supplying a slightly better one (my test one has the wrong transparency so there is whitespace around the bug)

demo-ui/app/uk/gov/ons/addressIndex/demoui/modules/VersionModule.scala
```diff
@@ -0,0 +1,57 @@
...
+  def apiVersion: String
+  def dataVersion: String
+}
+
+/**
+  * Singleton class to get the API version number and data epoch once only
+  * @param apiClient
+  */
+@Singleton
+class DemoUIAddressIndexVersionModule @Inject()(
```
Mironor: This is a very peculiar way of getting a version.
What I thought was more in the spirit of extracting it from api response while getting a normal response from the api (by input, by uprn, bulk).
If the version needs to be display on every page (thus in the main template), I would opt for a creation of a separate api endpoint `/version` that will return just that.

RichardSmithONS:
I did consider a separate endpoint, and also the possibility of returning the version information with the swagger definition (we currently don't have an endpoint for this but are intending to). Similarly there could be a general purpose system properties response (some REST servers have a built in /services resource). I will do an endpoint for just the version for now.

demo-ui/public/stylesheets/ons.css
```diff
@@ -655,11 +699,42 @@ div.taColumn4H{
...
+        display: inline-block;
+        padding-bottom:2px;
+        padding-top:2px;
+        margin-bottom:4px;
+        margin-top:4px;
+        margin-left:4px;
+    }
+
+    details{
+        min-width: 400px;
```
steve-thorne-ONS: The drop in width will create more address wrapping.

demo-ui/public/stylesheets/ons.css
```diff
@@ -655,11 +699,42 @@ div.taColumn4H{
...
+    }
+
+    details details div{
+        text-indent: 20px;
```
steve-thorne-ONS: No longer looks like the JIRA example with the text-indent change.

demo-ui/public/stylesheets/ons.css
```diff
@@ -655,11 +699,42 @@ div.taColumn4H{
...
+        font-size:0.6em
```
steve-thorne-ONS: The font size change corrects the wrapping problem but makes it very hard to read in a browser.
In danger of sacrificing demoUI usability to enable small screen compatibility. I think you need to run these changes past Taiwo since the UI was previously agreed.

RichardSmithONS: They are inside a _@media only screen_ block so only apply to small screen devices. Now that the Exeter index is fixed I can do a push and test on branch with phone and desktop.

demo-ui/app/uk/gov/ons/addressIndex/demoui/modules/VersionModule.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    Try(Await.result(
+      apiClient.versionQuery()
+        .map { resp: AddressResponseVersion =>
+          resp.dataVersion
+        }, 10 seconds)).getOrElse(exampleDataVersion)
+  }
+
+  val exampleApiVersion = "e1234b"
+  val exampleDataVersion = "39"
```
Mironor: May we replace those values by "not defined"

PR #216, opened 2017-05-16
Hopper Score Amendment #2 (JIRA 536)

RichardSmithONS:
This is the first  of 5 improvements to the Hopper score, better dealing with numbers in text fields. It has been tested on internal branch.


RichardSmithONS:
Will change the brackets as codacy requested for next time

PR #264, opened 2017-09-29
fix problem with duplicate postcode part (JIRA 611)

RichardSmithONS:
Small change to Tokens.scala and add new test


parsers/src/main/scala/uk/gov/ons/addressIndex/parsers/Tokens.scala
```diff
@@ -92,6 +93,7 @@ object Tokens {
     * @return Map with label -> concatenated tokens ready to be sent to the ES
     */
   def postTokenize(tokens: Map[String, String]): Map[String, String] = {
+    println(tokens)
```
steve-thorne-ONS: Should be removed.

PR #266, opened 2017-10-16
Changes to application conf and home controller for login

RichardSmithONS:
Old Gateway now working, some small changes required to policy (held in on-site GitLab not public GitHub) and to the  conf and  to  the  home  controller. It  now works, provided the old gateway's IP is specified in Jenkins. New gateway still not quite there.

### DIFF
--- a/demo-ui/conf/application.conf
+++ b/demo-ui/conf/application.conf
@@ -21,8 +21,9 @@ demoui {
   customErrorTest =  ${?ONS_AI_UI_CUSTOM_ERROR_TEST}
   customErrorProd = true
   customErrorProd =  ${?ONS_AI_UI_CUSTOM_ERROR_PROD}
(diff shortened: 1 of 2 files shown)


demo-ui/conf/application.conf
```diff
@@ -21,8 +21,9 @@ demoui {
   customErrorTest =  ${?ONS_AI_UI_CUSTOM_ERROR_TEST}
   customErrorProd = true
   customErrorProd =  ${?ONS_AI_UI_CUSTOM_ERROR_PROD}
-  gatewayURL = "https://apigw-in-d-02:9443"
- // gatewayURL = "https://10.50.14.22:9443"
+ // gatewayURL = "https://apigw-in-d-02:9443"
+  gatewayURL = "https://10.50.14.22:9443"
```
steve-thorne-ONS: Remove IP.

PR #305, opened 2018-03-29
Implemented historical and non-historical indexes.

steve-thorne-ONS:
Added Non-Historical index.


server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -165,16 +169,17 @@ class AddressIndexRepository @Inject()(
```
RichardSmithONS: Does postcode need to be able to use the historical or is it always current only?

steve-thorne-ONS: Yes it does. I missed that as part of the merge conflict resolution. I'll update.

PR #346, opened 2018-07-30
Performance tests added to address-index-api server

ajorpheus:
This change adds performance tests meant to test an instance of
address-index-api server.

These tests are ported over from registers-performance-tests repo.
The associated documentation shall be updated in subsequent changes.


ajorpheus:
Branch name does not confirm to convention. Closing.

PR #347, opened 2018-07-30
Performance tests added to address-index-api server

ajorpheus:
This change adds performance tests meant to test an instance of
address-index-api server.

These tests are ported over from registers-performance-tests repo.
The associated documentation shall be updated in subsequent changes.


ajorpheus:
Thanks @RichardSmithONS .  There is a provision of using the generic get request to pass in whatever address one wishes. At the time of this writing this is how **[address-index-api-dev-gatling](http://jen-download-01:8080/job/address-index-api-dev-gatling/configure)** is executed, with the following params:

` -DconcurrentUsers=100 -DBASE_URL=http://addressindex-api-dev.apps.cf1.ons.statistics.gov.uk/addresses?input=26%20FROG%20LANE%20TITCHFIELD%20PO144DU`

As long as the Address is url encoded, we can use whatever address we wish for GET /addresses endpoint.

PR #353, opened 2018-08-02
Corrected the list of cipher suites used by Gatling's AHC

ajorpheus:
This allows the gatling test to be run against an https endpoint. This
configuration might have to be amended later for other https endpoints.

At the time of this commit, this configuration was tested against
https://add??-??-dev.app.cf1 (actual endpoint elided)


ajorpheus:
> I was off sick when they did Cipher Suites at school. OK to merge.

:)

PR #355, opened 2018-08-06
Updated documentation and minor fixes to configuration vars

ajorpheus:
Renamed numOfConcurrentUsers to numOfRequestsPerSecond since from the
perspective of an end-user, that is what this number represents (From
the point of view of the actual gatling simulation itself, this param is
used for both number of concurrent users as well as concurret requests
per second)

Renamed configName to CONFIG_NAME to align with the other system
properties that the gatling tests accept.


ajorpheus:
@RichardSmithONS  Thank  you.

PR #361, opened 2018-08-13
Override universalBin task for server to include parser resources

ajorpheus:
This change allows the removal of the manual 'jar uf' command that is
currently invoked after building a zip package.

Before the server (address-index-api/server) can be deployed to Cloud
Foundry, a zip package is created using:
 $ sbt "project address-index-server" "universal:packageBin"

and the zip modified by using a command like so:
    $ jar uf server/target/universal/address-index-server.zip \
     -C parsers/src/main/resources libcrftagger-linux.so \
     -C parsers/src/main/resources addressCRFA.crfsuite \
     -C parsers/src/main/resources/input_pre_post_processing business \
     -C parsers/src/main/resources/input_pre_post_processing company \
     -C parsers/src/main/resources/input_pre_post_processing county \
     .....

Ideally, it is desired that this be done entirely by the build tool,
otherwise this step would have to be replicated every time a package has
to be made (either by devs or by Jenkins).

This is especially relevant for creating Jenkins Pipeline jobs where
this manual 'jar uf' task would have to be replicated.


ajorpheus:
I need to test this by delpoying via Jenkins to make sure the server package deploys fine on CF

PR #362, opened 2018-08-13
Override universalBin task for server to include parser resources

ajorpheus:
This change allows the removal of all the 'shell' steps needed after
construction of the zip package via SBT. These manual steps include:
    1. Modifying the zip using 'jar uf' to add parser resources (details
below)
    2. Creating a copy of the zip to allow using a predicatable name for
the Jenkins Cloud Foundry plugin to push the app bundle.

Removing the need for the shell steps will make this build more portable
and easily reproducible.

Note that this required upgrading sbt-native-packager from v.1.1.4 to
v1.3.3

Before the server (address-index-api/server) can be deployed to Cloud
Foundry, a zip package is created using:
 $ sbt "project address-index-server" "universal:packageBin"

and the zip modified by using a command like so:
    $ jar uf server/target/universal/address-index-server.zip \
     -C parsers/src/main/resources libcrftagger-linux.so \
     -C parsers/src/main/resources addressCRFA.crfsuite \
     -C parsers/src/main/resources/input_pre_post_processing business \
     -C parsers/src/main/resources/input_pre_post_processing company \
     -C parsers/src/main/resources/input_pre_post_processing county \
     .....

Ideally, it is desired that all steps required to create a distribution
be done entirely by the build tool, otherwise these steps would have to
be replicated every time a package has to be made (either by devs or by
Jenkins).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,9 +1,10 @@
 import com.iheart.sbtPlaySwagger.SwaggerPlugin.autoImport.swaggerDomainNameSpaces
 import com.typesafe.sbt.SbtNativePackager.autoImport.NativePackagerHelper._
+import com.typesafe.sbt.packager.universal.ZipHelper
 import com.typesafe.sbt.web.SbtWeb
 import play.sbt.PlayScala
 import play.sbt.routes.RoutesKeys._
-import sbt.Keys._
+import sbt.Keys.{mappings, _}
 import sbt.Resolver.{file => _, url => _}
 import sbt._
 import sbtassembly.AssemblyPlugin.autoImport._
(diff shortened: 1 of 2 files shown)


ajorpheus:
Tested on CF Dev deployment : http://jen-download-01:8080/job/address-index-api-dev/429/console). Note that this deployment was made from the branch of  this PR.

Finally, tested the actual endpoint via : https://addressindex-api-dev.apps.cf1.ons.statistics.gov.uk/addresses?input=26%20FROG%20LANE%20TITCHFIELD%20PO144DU

willis7:
Do the parser files need to be on the classpath? If so, wouldn't it be easier just to move them into the resources directory of the API? They would automatically get packaged and placed on the classpath.

ajorpheus:
Thanks @steve-thorne-ONS

PR #369, opened 2018-08-24
Introduce verbose parameter for all endpoints (1686)

RichardSmithONS:
main single match - verbose flag defaults to true and unwanted data must be discarded after scoring and thresholding
postcode - verbose flag defaults to true (might want to make this false), unwanted data items not gathered
UPRN - verbose flag defaults to true, unwanted data items not gathered
partial - verbose flag defaults to false, unwanted data items not gathered
for consistency partial now uses the same objects as the others rather than its own so the output is slightly different.
The UI has been changed to support verbose settings, and false is used when expanding relatives, everything else uses verbose true.
Unit tests have been updated but what I haven't done is double up everything for a verbose and concise version - will do a separate PR for this if needed and any other changes we decide upon (such as the default for postcode)


RichardSmithONS:
Consider making all defaults FALSE .. ?

PR #376, opened 2018-09-11
Create configuration and shell scripts for executing /bulk Gatling tests

ajorpheus:
Note that there are no changes to the scala code. The only changes
required were to add additional configuration to support the bulk
endpoint (for instance the json payload for POST request)


ajorpheus:
I have only tested this against a local AI instance. This PR must not be merged in until I have created a Jenkins job to confirm if this PR works as expected on-network.

PR #379, opened 2018-09-18
Adds static code analysis and coverage

ajorpheus:
This PR adds Cobertura, ScalaStyle and Scapegoat plugins to the build process


ajorpheus:
Hello Richard,

1. We are using the defaults for these tools for now, but these can be customised.
2.  Scapegoat has been added in this PR, I can look at Codacy in the next Sprint

PR #380, opened 2018-09-18
Replaces email id with a variable defined in Jenkins configuration

ajorpheus:
AI_EMAIL_LIST was created as a part of RITM0164506 for Registers under:
http://jen-download-01:8080/view/Registers/configure-readonly/

As of this commit, it only contains one email id, but can be amended
later to include the email ids of other AI devs


ajorpheus:
Yes, agreed. For that reason, I've only added my email id to the list for the time being till the builds are stable

PR #382, opened 2018-09-18
Creates a Jenkinsfile for the new Jenkins

ajorpheus:
Added a Jenkinsfile for new Jenkins

This commit creates a Jenkinsfile which can be used for dev branches on
the new Jenkins which has dockerized ephemeral slaves.

Things to note:
1. A test had to be ignored to allow the test stage to proceed. This has
been captured in REG-1985
2. The Gatling tests are limited to Open profile only ('destructive' /
ever increasing load)
3. At the moment, both ScapeGoat and ScalaStyle analysis are being done
4. Corbertura is being used to generate test code coverage reports


ajorpheus:
Thanks @RichardSmithONS

PR #397, opened 2018-10-05
Fix Static Code Analysis to show ScalaStyle Results

ajorpheus:
This commit:
 1. Upgrades scalastyle plugin version to the latest 1.0.0
 2. Removes scapeGoat static analysis since the reports generated from
it overwrite the ones created by Scalastyle after which neither are
accessible

### DIFF
--- a/Jenkinsfile-new
+++ b/Jenkinsfile-new
@@ -105,19 +105,6 @@ pipeline {
                         }
                     }
                 }
-                stage('Scapegoat') {
-                    agent { label 'build.sbt_0-13-13' }
-                    steps {
-                        unstash name: 'Checkout'
-                        colourText("info", "Running additional tests")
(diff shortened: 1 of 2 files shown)


ajorpheus:
Read more here : https://collaborate2.ons.gov.uk/jira/browse/REG-1984

ajorpheus:
Thanks Richard.

I am not certain why that error was being flagged by ScalaStyle, however (details are on REG-1984), there is one other class (AddressResponseAddress , Line 32) which has a trailing comma as well, but is not flagged by Scalastyle.

In any case, thanks to addition of another var in another change (https://github.com/ONSdigital/address-index-api/commit/e6d10c415d1001aeb717e2b5decba16e7bb870e2#diff-b6d65a3776be4e751ce131feb3af6233R352)  the fix is no longer needed.

PR #399, opened 2018-10-05
Move senstive address data to Gitlab

ajorpheus:
Move sensitive address data out of Github

The address payload jsons used for /bulk* POST requests has been moved
to Gitlab. Going forward, any sensitive data should be created in Gitlab.

The Gitlab repository has a directory called Gatling. Anything inside
there will be copied to "server/src/it/resources/uk/gov/ons/gatling/conf"
before the Gatling tests are run.


ajorpheus:
Looks good to me

PR #402, opened 2018-10-09
Fix default encoding being used for Jenkins causing a test failure

ajorpheus:
The default encoding being picked up by the JVM on jen-m Jenkins was
causing an error while reading a file containing an accented character.
The default encoding for Travis CI builds (and for local execution of
tests) was UTF-8, however for the Jenkins JVM it was not.

Instead of a code change to specify the encoding which might have impact
elsewhere, I chose to specify the encoding only for the specific SBT
task ('test') which obviates the need to evaluate the impact elswhere.

Task: REG-1985

### DIFF
--- a/Jenkinsfile-new
+++ b/Jenkinsfile-new
@@ -62,16 +62,13 @@ pipeline {
                     agent { label 'build.sbt_0-13-13' }
                     steps {
                         unstash name: 'Checkout'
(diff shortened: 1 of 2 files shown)


ajorpheus:
Looks good to me.

ajorpheus:
Great, thanks

PR #411, opened 2018-10-29
Error message improvements

RichardSmithONS:
This PR covers REG 2156 and 2157 (2158 to follow).

- Text of messages improved
- Redundant messages removed
- Unnecessary duplication removed
- Values (such as maxima) from the config file can now be inserted
- All messages now return an empty message appropriate to the type of response expected


model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/address/AddressResponseError.scala
```diff
@@ -54,160 +55,133 @@ object OffsetTooSmallAddressResponseError extends AddressResponseError(
...
 object FailedRequestToEsError extends AddressResponseError(
   code = 10,
-  message = "Request to Elasticsearch failed (check the API logs)"
+  message = "Request to Elasticsearch failed (see logs)"
 )
 
 object ApiKeyMissingError extends AddressResponseError(
   code = 11,
-  message = "The API key is missing"
+  message = "The API key is missing. It should be provided via the Authorization header."
```
steve-thorne-ONS: The API key is missing from the Authorization header.

PR #436, opened 2019-01-17
Performance improvement for typeahead (REG-2404)

RichardSmithONS:
Various index and query settings changes (such as edge-ngram min and max) were tested. The only one that yielded a noticeable improvement was to take the addresspostal=N filter out of the query and instead downgrade the Ns in post-processing in a similar way to  the startswith booster.

The effect is seen for the vaguer, shorter input strings where the ngram multimatch yields a high number of results. In this case the filter can add up to 300ms to the response time.

The downside of this is that some ponds etc. could creep back into the results. Testing is required to see how widespread this might be. If necessary the API could request a few extra results from ES then the list truncated to remove the downgraded Ns.


RichardSmithONS:
The skinny index now has the Ns removed, and the ES query now adds the filter only for the fat index. No results are now discarded or rescored in post-processing.

PR #454, opened 2019-04-08
FWMT-888, update to allow empty term on range filtering, test passed …

SteveScorfield:
…and outstanding query on partial


RichardSmithONS:
superseded by 455

PR #463, opened 2019-05-08
Add codecov

steve-thorne-ONS:
66% not so good.


RichardSmithONS:
Can be merged to develop, master will have to wait for a bit. Badge images are not always rendering, but this is a known problem with GitHub (various threads about it) .

PR #474, opened 2019-06-25
Query code cleanup 1232

RichardSmithONS:
Changes made to structure of "full match" query with Iva
Results already tested (further tickets for small improvements raised)
Once merged to develop will need another merge to active NISRA branch


RichardSmithONS:
> If we are removing the date query should we just remove the tests rather than ignore them?

Ticket FMWT-1087 exists to fully remove the date code, I have commented that 1232 has partly done this. Makes sense to get 1087 done sooner rather than later, though.

PR #481, opened 2019-08-02
NISRA set boosts and make them configurable

RichardSmithONS:
Works OK except some issues with partial "all" (equal playing fields) where some further investigation is needed into scoring oddities.

### DIFF
--- a/server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
+++ b/server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
@@ -3,6 +3,7 @@ package uk.gov.ons.addressIndex.server.modules
 import com.sksamuel.elastic4s.analyzers.CustomAnalyzer
 import com.sksamuel.elastic4s.http.ElasticDsl.{geoDistanceQuery, _}
 import com.sksamuel.elastic4s.http.HttpClient
+import com.sksamuel.elastic4s.http.search.SearchBodyBuilderFn
 import com.sksamuel.elastic4s.searches.queries.{BoolQueryDefinition, ConstantScoreDefinition, QueryDefinition}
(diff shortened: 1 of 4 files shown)


server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -734,10 +742,10 @@ class AddressIndexRepository @Inject()(conf: AddressIndexConfigModule,
           .minimumShouldMatch(queryParams.fallback.fallbackMinimumShouldMatch)
           .analyzer(CustomAnalyzer("welsh_split_synonyms_analyzer"))
           .boost(queryParams.fallback.fallbackLpiBoost),
-        matchQuery("nisra.nisraAll", normalizedInput)
+        matchQuery("nisra.nisraAll",  normalizedInput)
```
steve-thorne-ONS: Extra space?

server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -109,9 +116,10 @@ class AddressIndexRepository @Inject()(conf: AddressIndexConfigModule,
       case _ => Seq.empty
     }
 
+    // test with and without this - probably better not to have it was we have boost numbers
     val fromSourceQueryShould = args.fromsource match {
-      case "niboost" => Seq(termsQuery("fromSource","NI"))
-      case "ewboost" => Seq(termsQuery("fromSource","EW"))
+    //  case "niboost" => Seq(termsQuery("fromSource","NI"))
```
steve-thorne-ONS: Are these comments needed?

RichardSmithONS: The extra should query may or may not still be needed in addition to the boosts inside the multimatch (going without it for initial tests), so is commented rather than removed for the time being.

PR #483, opened 2019-08-08
API changes for secondary postcode sort feature (FWMT-1109)

RichardSmithONS:
Simple change to add the new invisible sort field (requires new index) to the postcode query.


RichardSmithONS:
> How much bigger is the index now? Will it impact performance? Is this a change that is useful for all clients or are we adding client sort code to the API?

The sort is only for the postcode results. It adds about 0.8GB to the fat index, and we expect to claw back some space on the thin index when the date fields are removed (there is a ticket for this). No impact on performance. Details of the data changes here https://github.com/ONSdigital/address-index-data/pull/137

PR #529, opened 2020-01-30
Add new fields to output

RichardSmithONS:
Not including logic to derive address and estab type


server/conf/application.conf
```diff
@@ -178,22 +178,23 @@ addressIndex {
...
+ //   uri = "search-ai-es-p-01-62qk73r6grdfk5nlefyi5tqyxe.eu-west-1.es.amazonaws.com"
     uri = ${?ONS_AI_API_ES_URI}
     // For GCP multiple ES clusters
     uriFullmatch = "localhost"
     uriFullmatch = ${?ONS_AI_API_ES_URI_FULLMATCH}
 
-//  port = "9200"
+      port = "9200"
+ //   port = "80"
     port = "80"
```
steve-thorne-ONS: Looks like 2 ports are live.

RichardSmithONS: Merge glitch - I have tidied it.

PR #533, opened 2020-03-13
Feature/fwmt 1917

steve-thorne-ONS:
It's a big one.


steve-thorne-ONS:
Agree with the highverbose comment. I missed this - probably because the class was created before the merge. It can be updated in the next code change along with the typo.

PR #559, opened 2020-06-26
Bugfixes for NISRA data and EQ / RH Endpoints

RichardSmithONS:
Tickets 2270 and 2300

### DIFF
--- a/server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
+++ b/server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
@@ -802,7 +802,7 @@ class AddressResponseAddressSpec extends WordSpec with Matchers {
 
     "create AddressByEqUprnResponse from Hybrid ES response for PAF" in {
       // Given
-      val hybrid = HybridAddress(givenPafEq.uprn, givenPafEq.uprn, Some(Seq(givenRelative)), Some(Seq(givenCrossRef)), Some("postcodeIn"), Some("postcodeOut"), Seq(givenNag), Seq(givenPafEq), Seq(givenNisra), 1, "classificationCode", "NA", "NA", "EW", "E", 0D, Seq() )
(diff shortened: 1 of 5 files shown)


server/test/uk/gov/ons/addressIndex/server/model/response/AddressResponseAddressSpec.scala
```diff
@@ -802,7 +802,7 @@ class AddressResponseAddressSpec extends WordSpec with Matchers {
 
     "create AddressByEqUprnResponse from Hybrid ES response for PAF" in {
       // Given
-      val hybrid = HybridAddress(givenPafEq.uprn, givenPafEq.uprn, Some(Seq(givenRelative)), Some(Seq(givenCrossRef)), Some("postcodeIn"), Some("postcodeOut"), Seq(givenNag), Seq(givenPafEq), Seq(givenNisra), 1, "classificationCode", "NA", "NA", "EW", "E", 0D, Seq() )
```
steve-thorne-ONS, line 805: I notice that ```givenNisra``` has been removed from a lot of tests. Does Nisra still have good test coverage?

RichardSmithONS: The fix for NISRA fallback relies on NISRA and non-NISRA being mutually exclusive, so there must not be both in the tests. We could add and extra NISRA only test.

steve-thorne-ONS: Ok. I think you should add that extra test.

RichardSmithONS: The existing tests on lines 891 and 1303 cover the NISRA cases. They didn't fail before because postTown and townName are both populated on the test NisraAddress object. The postTown field was removed from the real index due to being a duplicate of townName, but it still exists as an empty field on the object.

PR #560, opened 2020-07-07
Add new special bucketed postcode endpoint for eq

RichardSmithONS:
Ticket FWMT-2325

### DIFF
--- a/server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
+++ b/server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
@@ -285,10 +285,6 @@ class AddressIndexRepository @Inject()(conf: ConfigModule,
 }
 
   private def makeBucketQuery(args: BucketArgs): SearchRequest = {
-//    val postcodeFormatted: String = if (!args.postcode.contains(" ")) {
-//      val (postcodeStart, postcodeEnd) = args.postcode.splitAt(args.postcode.length() - 3)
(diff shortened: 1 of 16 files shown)


server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexRepository.scala
```diff
@@ -284,6 +284,42 @@ class AddressIndexRepository @Inject()(conf: ConfigModule,
     .limit(args.limit)
 }
 
+  private def makeBucketQuery(args: BucketArgs): SearchRequest = {
+//    val postcodeFormatted: String = if (!args.postcode.contains(" ")) {
+//      val (postcodeStart, postcodeEnd) = args.postcode.splitAt(args.postcode.length() - 3)
+//      (postcodeStart + " " + postcodeEnd).toUpperCase
+//    } else args.postcode.toUpperCase
```
steve-thorne-ONS, line 291: Needed?

RichardSmithONS: That space insertion code only works for a prevalidated full postcode but for this endpoint it can be 
- a full postcode copied from the groupedpostcode output (in which case it will already have the space)
- a wildcard (*)
- part of a postcode plus a wildcard (in which case the API should honour the pattern as supplied)

RichardSmithONS: I'll remove the code (and fix the readme for grouped postcode spotted today)

PR #573, opened 2020-09-16
fwmt-2437 added kamon instrumentation

gsmanderson:


RichardSmithONS (commented):
It should be OK, but we should really try a build with this branch on-prem to see if Artifactory is OK with the kamon libraries. What usually happens is it works second time.

RichardSmithONS:
Trying a build now on "dev"

RichardSmithONS:
Been running for 11 minutes, will leave it for a bit longer...

RichardSmithONS:
37 minutes and counting, but it is downloading stuff, lots of.

RichardSmithONS (approved):
Success!

PR #575, opened 2020-09-18
fwmt-2458 set nisra to none when verbose is false

gsmanderson:


RichardSmithONS (approved):
LGTM (the change, the original code is a bit of a hack, probably one of mine :-) )

PR #580, opened 2020-09-30
Add health check endpoint.

steve-thorne-ONS:


RichardSmithONS (approved):
Kudos for using HTTP code 418 :-)

PR #581, opened 2020-10-13
New Combo option for postcodes

RichardSmithONS:
Combined query is slower than the normal postcode one but much quicker than doing two. Very flexible for EQ and hopefully the performance hit will be acceptable.

### DIFF
--- a/model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/eq/AddressByEQPostcodeResponse.scala
+++ b/model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/eq/AddressByEQPostcodeResponse.scala
@@ -2,6 +2,7 @@ package uk.gov.ons.addressIndex.model.server.response.eq
 
 import play.api.libs.json.{Format, Json}
 import uk.gov.ons.addressIndex.model.server.response.address.AddressResponseAddressPostcodeEQ
+import uk.gov.ons.addressIndex.model.server.response.postcode.AddressResponsePostcodeGroup
 
 /**
   * Contains relevant, to the address request, data
@@ -19,6 +20,7 @@ import uk.gov.ons.addressIndex.model.server.response.address.AddressResponseAddr
   */
 case class AddressByEQPostcodeResponse(postcode: String,
                                        addresses: Seq[AddressResponseAddressPostcodeEQ],
+                                       postcodes: Seq[AddressResponsePostcodeGroup],
                                        filter: String,
                                        historical: Boolean,
                                        epoch: String,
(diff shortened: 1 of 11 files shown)


model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/eq/AddressByEQPostcodeResponse.scala
```diff
@@ -19,14 +20,16 @@ import uk.gov.ons.addressIndex.model.server.response.address.AddressResponseAddr
   */
 case class AddressByEQPostcodeResponse(postcode: String,
                                        addresses: Seq[AddressResponseAddressPostcodeEQ],
+                                       postcodes: Option[Seq[AddressResponsePostcodeGroup]],
```
gsmanderson, line 23: I would tend to favour just the Seq here as its either empty or not. Similar to option.

server/app/uk/gov/ons/addressIndex/server/controllers/EQController.scala
```diff
@@ -36,25 +36,14 @@ class EQController @Inject () (val controllerComponents: ControllerComponents,
               groupfullpostcodes: Option[String] = None
              ): Action[AnyContent] = Action async { implicit req =>
 
-    val groupFullPostcodesVal: Boolean = Try(groupfullpostcodes.getOrElse("false").toBoolean).toOption.getOrElse(false)
+ // groupfullpostocdes = yes, no, combo
```
gsmanderson, line 39: I would have looked at using an enum rather than a string for yes/no/combo. Just my personal choice  :) would  approach the enum using sealed case objects as shown here https://docs.scala-lang.org/overviews/scala-book/enumerations-pizza-class.html

model/src/main/scala/uk/gov/ons/addressIndex/model/server/response/eq/AddressByEQPostcodeResponse.scala
RichardSmithONS: It needs to be an option so that the JSON response where the combo is not selected is unchanged. Otherwise all the "regular" responses will contain _postcodes: []_ which is at best untidy and at worst a breaking change (if the UI is not optimally coded).

server/app/uk/gov/ons/addressIndex/server/controllers/EQController.scala
RichardSmithONS: I'll have a look at doing an enum

gsmanderson (approved):

PR #588, opened 2020-11-11
Rejig partial query and scoring script

RichardSmithONS:
Tested on-prem and solves a performance issue and the results appear to be fine after a bit of tuning.


gsmanderson (approved):
Looks good. Probably wouldn't have created the scriptText val as the string is only used once. 
I would have used the string directly when creating the partialScript val.

PR #598, opened 2021-01-08
fwmt-2918 - aux index scoring

gsmanderson:
…0 split


RichardSmithONS (approved):
This wasn't an easy one but I think this is the most pragmatic solution

PR #603, opened 2021-01-28
fwmt-3004 - adjusted for blank lpi location and removed aux only sort

gsmanderson:


RichardSmithONS (approved):
LGTM - agree use of Try is a bit lazy

PR #608, opened 2021-02-11
Enable access to restricted index via private key

RichardSmithONS:
Has been tested on-prem via a special user created in KIbana


gsmanderson (approved):
LGTM

PR #613, opened 2021-02-17
fwmt-2567 - increased a kamon value for http connection

gsmanderson:


MallikaM-ons (approved):
LGTM

RichardSmithONS (approved):
that's a lot of zeros, rather like our GCP bill :-)

PR #615, opened 2021-02-19
fwmt-2567 - increasing http connection lifetime as clients may hold c…

gsmanderson:
…onnection open


RichardSmithONS (approved):
7.7 months should be enough :-)

PR #625, opened 2021-04-20
Removed Docker and Artifactory settings.

steve-thorne-ONS:


RichardSmithONS (approved):
Yup, don't need those bits any more

PR #632, opened 2021-07-13
Change single match scale factor to match on-prem

RichardSmithONS:
ONS_AI_API_SCALE_FACTOR is for singles and we lowered this for CC (aux results)

ONS_AI_API_BULK_SCALE_FACTOR is the bulk setting and this is set to 23 

Whilst the JIRA ticket to tune the scale factor(s) is still outstanding, a figure of 21 has been working well on the single matches (mainly via UI) on-prem so we want to restore this to the default for GCP deployments.


steve-thorne-ONS (approved):
LGTM

PR #633, opened 2021-08-10
Update sbt in API project to newest that works with play 2.6

RichardSmithONS:
Swagger upgraded to V3 and on-demand class added.


steve-thorne-ONS (approved):
LGTM

PR #645, opened 2021-09-13
JDK Version change to avoid CRF load error

RichardSmithONS:
Test build deployed on initial-test-3


steve-thorne-ONS (approved):
Agreed as a temporary fix.

PR #646, opened 2021-09-17
Merge JDK fix to master

RichardSmithONS:
Current master code doesn't run, develop does. On-prem service uses feature/masteronprem (old sbt)


steve-thorne-ONS (approved):
LGTM

PR #657, opened 2022-02-21
Create new Epoch list endpoint

RichardSmithONS:
Slightly refactored from original class structure. Output slightly changed so description can be the release date of the epoch or something else like "Exeter test index"


server/app/uk/gov/ons/addressIndex/server/modules/AddressIndexVersionModule.scala
```diff
@@ -22,6 +23,27 @@ class AddressIndexVersionModule @Inject()
   val client: ElasticClient = elasticClientProvider.client
   val termsAndConditions = configProvider.config.termsAndConditionsLink
 
+//  lazy val epochList: List[String] = {
+//    List ("39","89","NA")
+//  }
+
+  lazy val epochDates: Map[String,String] = {
```
steve-thorne-ONS, line 30: Would this be more maintainable as a property file?
Up to you.

RichardSmithONS: When Felix and I discussed this we talked about trying to get the info from the OS website (a "scrape"), holding the descriptions in an external text file, or just putting a list in the code that would last for a year or so, and decided on option 3, being the simplest.

Another change considered was to dispense with the available epochs environment variable, and use the "real" list instead. This would be OK apart from when an epoch was loaded which should not be used (at the current time) and this does happen.

steve-thorne-ONS: Ok

steve-thorne-ONS (approved):
LGTM

PR #679, opened 2023-01-10
Boost scores of addresses that can receive post 

RichardSmithONS:
The extra boost is just enough to nudge the addressBasePostal=N addresses down the list, except where they really are a better match. Note that this does seem to fix the property shell issue noted in the recent NLP testing.

I have also removed NISRA fields from the queries (breaking a lot of tests) but the aux ones (tokens.xxxx) are still there.


steve-thorne-ONS (approved):
LGTM

PR #715, opened 2024-01-10
Switch off Circuit Breaker using env variable

RichardSmithONS:
Also swagger update missed from last PR


steve-thorne-ONS (approved):
LGTM

